PR #6, opened 2021-03-11
Uprev MobileCoin to ef5f81d

sugargoat:
Soundtrack of this PR: [Tomorrow is a Long Time, Bob Dylan](https://www.youtube.com/watch?v=UHG06Q45ziI)

### Motivation

Uprev the mobilecoin submodule to latest

### In this PR
* Uprev
* Fix up FogPubkeyResolver usage

### DIFF
--- a/libmobilecoin/src/transaction.rs
+++ b/libmobilecoin/src/transaction.rs
@@ -288,11 +288,16 @@ pub extern "C" fn mc_transaction_builder_create(
     fog_resolver: FfiOptRefPtr<McFogResolver>,
 ) -> FfiOptOwnedPtr<McTransactionBuilder> {
     ffi_boundary(|| {
-        let fog_resolver = fog_resolver
-            .as_ref()
-            .map_or_else(FogResolver::default, |fog_resolver| {
-                FogResolver::new(fog_resolver.0.clone(), &fog_resolver.1)
-            });
+        let fog_resolver =
+            fog_resolver
+                .as_ref()
+                .map_or_else(FogResolver::default, |fog_resolver| {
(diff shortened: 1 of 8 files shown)


libmobilecoin/src/transaction.rs
```diff
@@ -288,11 +288,16 @@ pub extern "C" fn mc_transaction_builder_create(
...
-            });
+        let fog_resolver =
+            fog_resolver
+                .as_ref()
+                .map_or_else(FogResolver::default, |fog_resolver| {
+                    // It is safe to add an expect here (which should never occur) because
+                    // fogReportUrl is already checked in mc_fog_resolver_add_report_response
+                    // to be convertible to FogUri
+                    FogResolver::new(fog_resolver.0.clone(), &fog_resolver.1)
+                        .expect("FogResolver could not be constructed from the provided materials")
```
eranrund, line 299: @kylefleming is this the desired behavior here?

kylefleming: Yes, this is fine.

Here's previous my comment on the internal repo PR where I requested the comment be added:
> Maybe let's add a comment saying why `expect` is safe here and should never occur. (The reason I think it's safe is that the fogReportUrl is already check in `mc_fog_resolver_add_report_response` to be convertible to `FogUri`.)

I'll add a caveat that this approach is slightly fragile. If the error type returned by `FogResolver` is ever expanded to include more errors, then this `expect` might no longer be safe to perform.

Ideally `FogReportResponses` uses `FogUri` as keys so that `FogResolver::new` isn't throwing an error for this.

kylefleming (approved):

eranrund (approved):

PR #9, opened 2021-03-11
Throw an error if the resolver cannot be created

voloshyn:
### In this PR
* Throw an error if FogResolver cannot be created


garbageslam (approved):
thank you!

sugargoat (approved):

PR #10, opened 2021-03-11
add some info-level logging when we successfully write a block

garbageslam:


sugargoat (approved):

eranrund (approved):

fog/ingest/server/src/controller.rs
```diff
@@ -491,6 +491,11 @@ where
                         self.new_egress_key(&mut state).expect("Failure to rotate egress key after we can't publish data isn't recoverable, the RNGs would have gaps that the clients can't deal with");
                     } else {
                         // We won the race to publish this block
+                        log::info!(
+                            self.logger,
+                            "Succeeded writing block {} to the database",
```
eranrund, line 496: Might be useful to include the key it is associated with.

PR #11, opened 2021-03-11
Fog 393 view server fixes

garbageslam:
This is supposed to make it so that view server does not get stuck on blocks that fog never promised to scan.

The main change is in the fog-view block tracker. To support the new idea, we make it not use ingestable ranges, but track ingress keys directly, and load blocks from database based on ingress key rather than ingest invocation id.

There are a couple additional places we try to unwind assumptions it was making like, every block leads to at least one ETxOutRecord, which isn't strictly speaking true. There is a code path in ingest enclave where a non-conforming client can send data that leads to no ETxOutRecord being produced, if the plaintext of encrypted fog hint isn't a valid ristretto.

Because it changes the DB API, a lot of unit tests need to be fixed up after this change. A bunch more need to be added for the new behavior as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,16 +863,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "displaydoc"
-version = "0.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "proc-macro2 1.0.24 (registry+https://github.com/rust-lang/crates.io-index)",
- "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
-name = "displaydoc"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
@@ -2622,7 +2612,7 @@ dependencies = [
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "crc 1.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "curve25519-dalek 3.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-account-keys 1.0.1-pre1",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
@@ -2642,7 +2632,7 @@ dependencies = [
  "aead 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "digest 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
  "mc-crypto-noise 1.0.1-pre1",
@@ -2680,7 +2670,7 @@ dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "digest 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex_fmt 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mbedtls 0.5.1 (git+https://github.com/mobilecoinofficial/rust-mbedtls.git?tag=mc-0.3)",
@@ -2704,7 +2694,7 @@ dependencies = [
 name = "mc-attest-enclave-api"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-attest-ake 1.0.1-pre1",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-noise 1.0.1-pre1",
@@ -2791,7 +2781,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "aes-gcm 0.6.0 (git+https://github.com/mobilecoinofficial/AEADs?rev=d1a8517d3dd867ed9c5794002add67992a42f6aa)",
  "cookie 0.14.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
  "mc-attest-ake 1.0.1-pre1",
  "mc-attest-api 1.0.1-pre1",
@@ -2911,7 +2901,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "base64 0.12.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "fs_extra 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
@@ -3129,7 +3119,7 @@ dependencies = [
 name = "mc-crypto-x509-utils"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-crypto-keys 1.0.1-pre1",
  "pem 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "x509-signature 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -3175,7 +3165,7 @@ dependencies = [
 name = "mc-fog-report-connection"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
  "mc-account-keys 1.0.1-pre1",
  "mc-attest-core 1.0.1-pre1",
@@ -3192,7 +3182,7 @@ dependencies = [
 name = "mc-fog-report-validation"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-account-keys 1.0.1-pre1",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
@@ -3215,7 +3205,7 @@ dependencies = [
 name = "mc-fog-sig"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-account-keys 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
  "mc-crypto-x509-utils 1.0.1-pre1",
@@ -3239,7 +3229,7 @@ dependencies = [
 name = "mc-fog-sig-report"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-digestible-signature 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
@@ -3313,7 +3303,7 @@ dependencies = [
  "aes-gcm 0.6.0 (git+https://github.com/mobilecoinofficial/AEADs?rev=d1a8517d3dd867ed9c5794002add67992a42f6aa)",
  "blake2 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam-channel 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
  "hex_fmt 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -3455,7 +3445,7 @@ dependencies = [
 name = "mc-sgx-css"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha2 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -3477,7 +3467,7 @@ dependencies = [
 name = "mc-sgx-report-cache-api"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-attest-core 1.0.1-pre1",
  "mc-attest-enclave-api 1.0.1-pre1",
  "mc-util-serial 1.0.1-pre1",
@@ -3488,7 +3478,7 @@ dependencies = [
 name = "mc-sgx-report-cache-untrusted"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-attest-core 1.0.1-pre1",
  "mc-attest-enclave-api 1.0.1-pre1",
  "mc-attest-net 1.0.1-pre1",
@@ -3538,7 +3528,7 @@ dependencies = [
  "blake2 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "bulletproofs 2.0.0 (git+https://github.com/dalek-cryptography/bulletproofs?rev=464acb72432ccc6c5131397fd811a5f01ec2454b)",
  "curve25519-dalek 3.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "generic-array 0.14.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex_fmt 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -3600,7 +3590,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "cargo_metadata 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mbedtls 0.5.1 (git+https://github.com/mobilecoinofficial/rust-mbedtls.git?tag=mc-0.3)",
  "mbedtls-sys-auto 2.18.1 (git+https://github.com/mobilecoinofficial/rust-mbedtls.git?tag=mc-0.3)",
  "mc-sgx-css 1.0.1-pre1",
@@ -3639,7 +3629,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "cc 1.0.66 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-util-build-script 1.0.1-pre1",
  "pkg-config 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -3650,7 +3640,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "base64 0.12.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "binascii 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-util-repr-bytes 1.0.1-pre1",
  "serde 1.0.118 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -3693,7 +3683,7 @@ version = "1.0.1-pre1"
 dependencies = [
  "base64 0.12.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "cookie 0.14.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -3741,7 +3731,7 @@ dependencies = [
 name = "mc-util-lmdb"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lmdb-rkv 0.14.0 (git+https://github.com/mozilla/lmdb-rs?rev=df1c2f5)",
  "mc-util-serial 1.0.1-pre1",
  "prost 0.6.1 (git+https://github.com/danburkert/prost?rev=6113789f70b69709820becba4242824b4fb3ffec)",
@@ -3812,7 +3802,7 @@ name = "mc-util-uri"
 version = "1.0.1-pre1"
 dependencies = [
  "base64 0.12.3 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ed25519 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-common 1.0.1-pre1",
@@ -3827,7 +3817,7 @@ dependencies = [
 name = "mc-watcher"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "grpcio 0.6.0 (git+https://github.com/jcape/grpc-rs?rev=2ad042e9e65ecb664a60e034d0a899a8760d81d3)",
  "hex 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "lmdb-rkv 0.14.0 (git+https://github.com/mozilla/lmdb-rs?rev=df1c2f5)",
@@ -3859,7 +3849,7 @@ dependencies = [
 name = "mc-watcher-api"
 version = "1.0.1-pre1"
 dependencies = [
- "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.118 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -5869,7 +5859,6 @@ dependencies = [
 "checksum dirs 2.0.2 (registry+https://github.com/rust-lang/crates.io-index)" = "13aea89a5c93364a98e9b37b2fa237effbb694d5cfe01c5b70941f7eb087d5e3"
 "checksum dirs-sys 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)" = "afa0b23de8fd801745c471deffa6e12d248f962c9fd4b4c33787b055599bde7b"
 "checksum discard 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)" = "212d0f5754cb6769937f4501cc0e67f4f4483c8d2c3e1e922ee9edbe4ab4c7c0"
-"checksum displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)" = "adc2ab4d5a16117f9029e9a6b5e4e79f4c67f6519bc134210d4d4a04ba31f41b"
 "checksum displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)" = "f98e1bbbcbb1b2646d6e900da42974b641eb784851adcd4f0ae4c07c8bc7b42b"
 "checksum downcast 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)" = "4bb454f0228b18c7f4c3b0ebbee346ed9c52e7443b0999cd543ff3571205701d"
 "checksum dtoa 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)" = "4358a9e11b9a09cf52383b451b49a169e8d797b68aa02301ff586d70d9661ea3"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,6 +863,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "displaydoc"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "proc-macro2 1.0.24 (registry+https://github.com/rust-lang/crates.io-index)",
+ "quote 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "displaydoc"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
@@ -1064,6 +1074,7 @@ dependencies = [
 name = "fog-ingest-client"
 version = "0.13.0"
 dependencies = [
+ "binascii 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "fog-api 0.13.0",
  "fog-types 0.13.0",
@@ -1868,6 +1879,7 @@ dependencies = [
  "mc-attest-net 1.0.1-pre1",
  "mc-common 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
+ "mc-crypto-x509-test-vectors 1.0.1-pre1",
  "mc-sgx-report-cache-untrusted 1.0.1-pre1",
  "mc-transaction-core 1.0.1-pre1",
  "mc-util-encodings 1.0.1-pre1",
@@ -1878,6 +1890,7 @@ dependencies = [
  "mc-util-serial 1.0.1-pre1",
  "mc-util-test-helper 1.0.1-pre1",
  "mc-util-uri 1.0.1-pre1",
+ "pem 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.7.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand_core 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.118 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -2609,7 +2622,7 @@ dependencies = [
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "crc 1.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "curve25519-dalek 3.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "mc-account-keys 1.0.1-pre1",
  "mc-attest-core 1.0.1-pre1",
  "mc-crypto-keys 1.0.1-pre1",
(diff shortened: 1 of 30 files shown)


tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -414,17 +414,6 @@ def run(self):
...
-        # Report a missed block range from 0 to 1. This is needed after FOG-337 and should not be needed after FOG-393
-        cmd = ' '.join([
-            f'exec {FOG_PROJECT_DIR}/{target_dir(self.release)}/fog_ingest_client',
-            f'--uri insecure-fog-ingest://localhost:{BASE_INGEST_CLIENT_PORT}',
-            f'report-missed-block-range',
-            f'--start 0',
-            f'--end 1',
-        ])
-        print(cmd)
-        result = subprocess.check_output(cmd, shell=True)
```
garbageslam, line 426: removing this is one of the main points of the PR

fog/view/server/src/server.rs
```diff
@@ -527,7 +548,8 @@ where
     // So, ask the database.
     // If it doesn't work, hopefully it will work next time.
     fn get_block_signature_timestamp_for_block_count(&self, block_count: u64) -> Option<u64> {
-        if block_count == 0 {
+        // The origin block has no block signature and hence no timestamp
+        if block_count <= 1 {
```
garbageslam, line 552: this was very spammy when running conformance tests, so i want to fix it

fog/recovery_db_iface/src/types.rs
```diff
@@ -5,8 +5,82 @@
...
+            .last_scanned_block
+            .map(|x| x + 1)
+            .unwrap_or(self.status.start_block);
+        if self.status.covers_block_index(candidate) {
+            Some(candidate)
+        } else {
+            None
+        }
+    }
+}
```
garbageslam, line 82: in the future, we should use this function in ingest instead of duping this logic

it's not used in this revision though I think, I ended up not using it. we could remove it for now

fog/view/server/src/block_tracker.rs
```diff
@@ -72,14 +71,33 @@ impl BlockTracker {
...
+    ///   If there are no ingress keys at all right now, then this is also the highest
+    ///   fully-processed block. It is assumed in the algorithm that any new keys will
+    ///   appear with start block after this number.
+    /// * ingress_keys:
+    ///   IngressPublicKeyRecord's that exist in the database right now.
+    ///   This indicates their start block, their last-scanned block, their expiry block,
+    ///   and whether they are retired.
+    /// * missing_block_ranges:
+    ///   Any manually entered missing block ranges.
+    ///
```
garbageslam, line 89: I need to add a code comment explaining the return value:

The idea is, we return what the highest fully processed block count is, and, if an Ingress key is what is blocking further progress, a copy of that record. So that if we are blocked on the same thing for a long time we can log

fog/view/server/src/db_fetcher.rs
```diff
@@ -268,60 +268,34 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> DbFetcherThread<DB> {
...
         log::trace!(
             self.logger,
             "load_block_data next_blocks: {:?}",
-            next_block_index_per_invocation_id
+            next_block_index_per_ingress_key
         );
 
-        for (ingest_invocation_id, block_index) in next_block_index_per_invocation_id.into_iter() {
-            // Do not attempt to fetch blocks explicitly marked as missing.
-            if BlockTracker::is_missing_block(&missing_block_ranges, block_index) {
```
garbageslam, line 289: I removed this because I'm worried it can cause a bug

Suppose that an ingest invocation id is chugging along, 0 ... 20

Then, by mistake perhaps, we report 5..10 as a missed block range

Then, the enclave will never load blocks 5..10 for this invocation id, so a bunch of users will have gaps in their RNGs.

garbageslam: i think its better to just load these blocks if they are in the database

fog/sql_recovery_db/src/lib.rs
```diff
@@ -241,6 +242,57 @@ impl RecoveryDb for SqlRecoveryDb {
...
+                        key: ingress_key,
+                        status,
+                        last_scanned_block: maybe_index.map(|x| x as u64),
+                    })
+                },
+            )
+            .collect::<Result<Vec<IngressPublicKeyRecord>, Error>>()?;
+
+        // Sort them as described in get_ingress_key_records trait comments
+        key_records.sort_unstable_by(|a, b| -> Ordering {
```
eranrund, line 286: With the subquery in place it might be possible to have the sql server do the ordering.
There's some relevant stuff in https://docs.diesel.rs/diesel/query_dsl/trait.QueryDsl.html#method.order and https://github.com/diesel-rs/diesel/issues/210

I can try doing it as well.
If you want to not spend time on this then at least wrap this whole thing in a read-only transaction (https://docs.diesel.rs/diesel/pg/struct.TransactionBuilder.html#method.read_only)

fog/sql_recovery_db/src/lib.rs
```diff
@@ -241,6 +242,57 @@ impl RecoveryDb for SqlRecoveryDb {
...
+            .iter()
+            .map(
+                |ingress_key_data| -> Result<IngressPublicKeyRecord, Error> {
+                    let status = IngressPublicKeyStatus {
+                        start_block: key_records[0].start_block as u64,
+                        pubkey_expiry: key_records[0].pubkey_expiry as u64,
+                        retired: key_records[0].retired,
+                    };
+
+                    // Look in ingested_blocks table for last scanned block
```
eranrund, line 266: Its probably better to do this with a subquery to cut down on roundtrips to the SQL server (and not make this an N+1 problem). There's an example [here](https://github.com/mobilecoinfoundation/fog/blob/master/fog/sql_recovery_db/src/lib.rs#L300-L302)

fog/sql_recovery_db/src/lib.rs
```diff
@@ -665,30 +717,36 @@ impl RecoveryDb for SqlRecoveryDb {
     /// Get any ETxOutRecords produced by a given IngestInvocationId for a given block index.
```
eranrund, line 717:
```suggestion
    /// Get any ETxOutRecords produced by a given ingress key for a given block index.
```

fog/view/server/src/block_tracker.rs
```diff
@@ -1,69 +1,68 @@
...
     pub fn new(logger: Logger) -> Self {
         Self {
-            processed_block_per_invocation: HashMap::default(),
+            processed_block_per_ingress_key: HashMap::default(),
             last_highest_processed_block_count: 0,
             logger,
         }
     }
 
     // Given a list of ingestable ranges and the current state, calculate which block
```
eranrund, line 30:
```suggestion
    // Given a list of ingress keys and the current state, calculate which block
```

fog/view/server/src/block_tracker.rs
```diff
@@ -1,69 +1,68 @@
...
                 // next one can be provided by it, and if so add it to the list of next blocks we
                 // would like to process.
                 let next_block = last_processed_block + 1;
-                if ingestable_range.can_provide_block(next_block) {
-                    next_blocks.insert(ingestable_range.id, next_block);
+                if rec.status.covers_block_index(next_block) {
+                    next_blocks.insert(rec.key, next_block);
                 }
             } else {
                 // No block has been processed for this invocation id, so the next block is the
```
eranrund, line 48:
```suggestion
                // No block has been processed for this ingress key, so the next block is the
```

fog/view/server/src/block_tracker.rs
```diff
@@ -72,14 +71,33 @@ impl BlockTracker {
 
     /// Given a list of ingestable ranges, missing blocks and current state, calculate the highest
```
eranrund, line 72:
```suggestion
    /// Given a list of ingress keys, missing blocks and current state, calculate the highest
```

fog/view/server/src/block_tracker.rs
```diff
@@ -72,14 +71,33 @@ impl BlockTracker {
...
     /// processed block count number. The highest processed block count number is the block count
-    /// for which we know we have loaded all available data.
+    /// for which we know we have loaded all required data, so the users can potentially compute
+    /// their balance up to this block without missing any transactions.
+    ///
+    /// Arguments:
+    /// * highest_known_block_index:
+    ///   The highest block index known to have appeared in the blockchain.
+    ///   If there are no ingress keys at all right now, then this is also the highest
```
eranrund, line 80: Isn't the value of this coming from the sql database?
If so, how would there ever be a value for it if there are no ingress keys?

fog/view/server/src/db_fetcher.rs
```diff
@@ -214,25 +218,21 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> DbFetcherThread<DB> {
         }
     }
 
-    /// Sync ingestable ranges from the database. This allows us to learn which ingest invocations
+    /// Sync ingress key records from the database. This allows us to learn which ingress keys
     /// are currently alive, which block ranges they are able to cover, and which blocks have they
     /// ingested so far.
-    fn load_ingestable_ranges(&self) {
+    fn load_ingress_keys(&self) {
         let _metrics_timer = counters::LOAD_INGESTABLE_RANGES_TIME.start_timer();
```
eranrund, line 225: This metric needs to be renamed.

fog/view/server/src/server.rs
```diff
@@ -343,6 +344,10 @@ where
     /// Keeps track of which blocks we have fed into the enclave.
     enclave_block_tracker: BlockTracker,
 
+    /// Keeps track ago we weren't blocked from making progress
```
eranrund, line 347: Missing a word or two?

fog/view/server/src/server.rs
```diff
@@ -397,21 +403,36 @@ where
...
-            .highest_fully_processed_block_count(&ingestable_ranges, &missing_block_ranges);
-        shared_state.highest_processed_block_count = highest_processed_block_count;
+        if shared_state.highest_processed_block_count != highest_processed_block_count {
+            shared_state.highest_processed_block_count = highest_processed_block_count;
+            self.last_unblocked_at = Instant::now();
+        } else if self.last_unblocked_at.elapsed() >= Duration::from_secs(60) {
+            if let Some(reason_we_stopped) = reason_we_stopped {
+                log::warn!(self.logger, "We seem to be stuck at highest_processed_block_count = {} for at least a minute... we are blocked on an ingress key making progress: {:?}", highest_processed_block_count, reason_we_stopped);
+            } else {
+                log::info!(self.logger, "We seem to be stuck at highest_processed_block_count = {} for at least a minute... we have processed all blocks known to the recovery database", highest_processed_block_count);
```
eranrund, line 431: Wouldn't this log once a minute for as long as there is no activity in the network?

tools/fog-local-network/fog_conformance_tests.py
eranrund: Should be removed from the local network script as well I think

fog/view/server/src/db_fetcher.rs
```diff
@@ -268,60 +268,34 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> DbFetcherThread<DB> {
...
-                Ok(tx_outs) => {
-                    let num_tx_outs = tx_outs.len();
-
-                    // NOTE: This makes a very nuanced and important assumption, which is that
-                    // ingest always produces data ETxOutRecords for each block it has processed,
-                    // EVEN if it actually found NO matches.
-                    // Based on that assumption, tx_outs will be empty only when ingest has not yet
-                    // ingested the block (and wrote the results into the database).
-                    if !tx_outs.is_empty() {
+                Ok(maybe_tx_outs) => {
```
eranrund, line 290: I think this will be more nicely written as 3 clauses in the match block:
```
Ok(Some(tx_outs)) => ...
Ok(None) => ...
Err(err) => ...
```

eranrund (commented):
Looks great so far! It is the type of code that is prone to off-by-one/edge case issues so the more unit tests we can get, the more confident we can be in it. I haven't noticed anything obviously wrong with it.

fog/view/server/src/server.rs
garbageslam: yeah, is that bad? i guess we could just remove this line

fog/view/server/src/db_fetcher.rs
garbageslam: oh yeah, that is better

fog/sql_recovery_db/src/lib.rs
garbageslam: when i started the PR i thought i would need this ordering but in the end nothing relies on it, we could probably punt on it

fog/view/server/src/server.rs
eranrund: I am not sure. Its nice to see things are still alive...
@jgreat @joekottke any thoughts on how much noise is too much noise? Context here is a log message that happens once a minute when the network is inactive. The log message lets us know fog-view is still alive but doesn't think it has anything to work on.

fog/sql_recovery_db/src/lib.rs
eranrund: If the ordering isn't needed then I would remove it and the mention in the documentation. That would make the transition from N+1 to 1 query with a subquery more trivial.

fog/view/server/src/block_tracker.rs
garbageslam: you're right, this is kind of a silly way to do this

this function should probably figure out the highest known block index based on what it sees in the ingress keys record set

eranrund: Maybe in the ingress keys list + missing blocks list...?

fog/view/server/src/server.rs
jayzalowitz: @eranrund, depends on your budget for datadog. Ideally I only want things I should actually react to as a normal engineer. Example: Give one alert for when network isnt connecting as an error log, but info logs should be responding with this heartbeat when its not working, this way I can debug better when things are wrong, but dont have to pay 10000/mo for the 10m log lines this could leave on datadog..

fog/view/server/src/db_fetcher.rs
```diff
@@ -495,94 +525,151 @@ mod tests {
...
+            blocks_and_records_40_50.push((block, records));
         }
 
         sleep(Duration::from_secs(1)); // Supposedly enough time for at least some blocks to get picked up.
 
         assert!(db_fetcher.get_pending_fetched_records().is_empty());
 
-        // Report a missing block range.
-        db.report_missed_block_range(&BlockRange::new(20, 30))
+        // TODO The last scanned block index gets updated even though we have a hole.
```
eranrund, line 566: @garbageslam I wonder what your thoughts on this are. There is a weird case where if somehow some blocks were skipped, `last_scanned_block` and `highest_known_block_index` would still go up. I think this is fine but wanted to check with you since this differs from the previous behavior, I think

fog/view/server/src/db_fetcher.rs
```diff
@@ -495,94 +525,151 @@ mod tests {
...
+                    retired: false,
+                },
+                last_scanned_block: Some(49),
+            }]
+        );
+        assert_eq!(highest_known_block_index, 49);
+        assert!(missing_block_ranges.is_empty());
+
+        // Report a missing block range. This should have no effect.
+        db.report_missed_block_range(&BlockRange::new(30, 40))
```
eranrund, line 586: The current version doesn't skip blocks that overlap with missed block ranges, instead it is only the pubkey expiry that controls stopping scanning for a given key. 

I want to check my understanding, @garbageslam is this flow correct?

1) We have an active ingress key with pubkey_expiry = 100, and it has scanned all the way to block 50.
2) Everything crashes and by the time we restarted, the network moved to block 75.
3) A new ingress key is created with start=75
4) We report missed block range 50-100, since we did advertise that we will scan until block 100 but failed
5) We set retired=true for the first ingress key, and this causes the view server to stop trying to load blocks from it

eranrund: Hm, I'm looking at https://github.com/garbageslam/fog/blob/1d924ff2a0663b432fbee854915c65ee02fb3246/fog/view/server/src/block_tracker.rs#L30-L32 and it looks like it will keep returning block `51` is the next one even if we are retired, because it is lower than the pubkey expiry. @garbageslam what are your thoughts on that?

eranrund:
@garbageslam some notes about tests:
* db_fetcher appears to have a flaky test `basic_single_ingress_key`
* all the `highest_fully_processed_block_` tests in `block_fetcher.rs` need to be reviewed. The addition of `highest_known_block_index` means there are more cases to check for.

garbageslam:
@eranrund as I'm working through these tests, I'm hitting some situations where, neither choice really seems ideal.

In this test for instance `next_blocks_single_key_retired_hasnt_scanned`:

```
        let mut block_tracker = BlockTracker::new(logger);

        let key = CompressedRistrettoPublic::from_random(&mut rng); 
        let rec = IngressPublicKeyRecord {
            key,
            status: IngressPublicKeyStatus {
                start_block: 123,
                pubkey_expiry: 173,
                retired: true,
            },
            last_scanned_block: None,
        };

        // This is the expected state because, even though the key is retired,
        // we promised to scan from 123 to 173. So the next thing we need is 123,
        // until we 
        // unless those blocks are declared missed.
        let expected_state = HashMap::from_iter(vec![(key, 123)]);
```

The code previously had "expected_state" = empty, but I think it should be, we need to load block 123

Because retired doesn't mean "lost" or "missing", it just means the operator is trying to shut down this key. But we already published it so it has been used and we have to scan with it or report missing blocks.

As I'm working through these tests, I'm thinking that, the best thing may be to fold into this PR a change I talked about and wanted to stage for later -- make the ingress keys have a boolean flag "lost" in addition to "retired", and make this the only pathway to report missed blocks to the DB. When a key is lost, the range "last_scanned" -> "pubkey_expiry" is missed. Lost means, there are no longer any blocks coming for this key. Then, next_blocks would take into account this field.

I think this would simplify the logic in the block tracker, so it may make it easier to develop this PR and develop a good body of tests for it and build confidence in the change. But it also increases the scope of the PR. But that's the route I'm thinking to take right now. LMK what you think

eranrund:
@garbageslam I think that sounds reasonable. What would be the 1-2 line description of `retired` in that case? When do we set it to true and what does that represent?

WIth regards to next_blocks() returning the start_block. The idea of it returning last_scanned_block and not start_block was to not have the view server (inside db_fetcher) try and fetch a block that was not scanned yet. If it [defaults](https://github.com/garbageslam/fog/blob/16017a2aa9b232b070a62ac5ea887734ace2c004/fog/view/server/src/block_tracker.rs#L53) to the start_block, then [this](https://github.com/garbageslam/fog/blob/16017a2aa9b232b070a62ac5ea887734ace2c004/fog/view/server/src/db_fetcher.rs#L273-L287) will repeatedly try to load it, even if we know (by last scanned block) that it never got written which is suboptimal.

garbageslam:
when retired is set, it means we are trying to retire -- we are no longer publishing reports for the key. but we still may have the key and we may still be scanning the last few blocks with it. things may get retired during "normal operation" of the system, like if we do an enclave upgrade. retired doesn't mean the users need to download blocks.

lost means, any blocks we were supposed to scan that we didn't scan, are missed and need to be downloaded by the user.
lost shouldn't happen during normal operation of the system, it happens if all the replicas crash

in both cases they are only set manually

garbageslam:
that is more than 1 line, i will work on it :)

eranrund:
I see, thank you for the explanation. So would it be correct to say that we are retiring while `last_scanned_block < (pubkey_expiry - 1)` and are actually retired when `last_scanned_block == (pubkey_expiry - 1)`? This is based on `pubkey_expiry` been the first block index that would not get scanned.

garbageslam:
yes exactly

eranrund:
Maybe not in this PR, but it seems to me that we should move from two booleans columns (`retired` and `lost`) into a `status` column that could be one of the following: `pending` (maybe? before anything got scanned. maybe not needed), `active`, `lost`, `retiring`, `retired`.

fog/view/server/src/server.rs
jgreat: If this isn't an error condition, I'd make this a debug message. Useful logging is usually about events in the system.

This kind of information we should be able to collect via metrics. If we see an abnormal pattern in blocks processes or a metric that shows which block we are on that we can compare with the current block on the chain, we can alert off of that.

fog/recovery_db_iface/src/lib.rs
```diff
@@ -210,16 +208,36 @@ pub trait RecoveryDb {
...
+
+    /// Get the invocation id that published this block with this key.
+    ///
+    /// Note: This is only used by TESTS right now, but it is important to be able to test this
+    ///
+    /// Arguments:
+    /// * ingress_key: The ingress key we are interested in
+    /// * block_index: the blcok we are interested in
+    ///
+    /// REturns:
```
eranrund, line 232:
```suggestion
    /// Returns:
```

fog/recovery_db_iface/src/types.rs
```diff
@@ -5,8 +5,87 @@
...
+    /// The first block that fog promises to scan with this key after publishing it.
+    /// This should be the latest block that existed before we published it (or, a block close to but before that)
+    pub start_block: u64,
+    /// The largest pubkey expiry value that we have ever published for this key.
+    /// If less than start_block, it means we have never published this key.
+    pub pubkey_expiry: u64,
+    /// Whether this key is retiring / retired.
+    /// When a key is retired, we stop publishing reports about it.
+    pub retired: bool,
+    /// Whether this key is lost
```
eranrund, line 23:
```suggestion
    /// Whether this key is lost.
```

fog/sql_recovery_db/src/lib.rs
```diff
@@ -1412,10 +1509,10 @@ mod tests {
         db.decommission_ingest_invocation(&invoc_ids[1]).unwrap();
 
         // Add two missing block records.
-        db.report_missed_block_range(&BlockRange::new(10, 20))
-            .unwrap();
-        db.report_missed_block_range(&BlockRange::new(30, 40))
-            .unwrap();
+        // db.report_missed_block_range(&BlockRange::new(10, 20))
```
eranrund, line 1512: Leftovers. Unless replaced by a different test, we should include reporting of a lost key and see that it translates to a missed blocks event.

fog/test_infra/src/db_tests.rs
```diff
@@ -169,66 +166,24 @@ pub fn recovery_db_smoke_tests_new_apis<DB: RecoveryDb>(
...
-    assert_eq!(
-        user_events,
-        vec![FogUserEvent::MissingBlocks(ok_block_range.clone())]
-    );
+pub fn recovery_db_missed_blocks_reporting(
+    rng: &mut (impl RngCore + CryptoRng),
+    db: &impl RecoveryDb,
+) {
+    let ingress_key = CompressedRistrettoPublic::from(RistrettoPublic::from_random(rng));
+    db.new_ingress_key(&ingress_key, 0).unwrap();
```
eranrund, line 174: I would expand this test to see what happens if we have overlapping ingress keys.
Its unlikely, but we might run into it while trying to deploy a new version and somehow losing the entire old and new clusters.

fog/test_infra/src/db_tests.rs
```diff
@@ -169,66 +166,24 @@ pub fn recovery_db_smoke_tests_new_apis<DB: RecoveryDb>(
...
 }
 
 // Basic tests that rng records decommissioning works as expected
 pub fn recovery_db_rng_records_decommissioning<DB: RecoveryDb>(
-    mut rng: &mut (impl RngCore + CryptoRng),
+    rng: &mut (impl RngCore + CryptoRng),
     db: &DB,
 ) {
-    let ingress_key = CompressedRistrettoPublic::from(RistrettoPublic::from_random(&mut rng));
+    let ingress_key = CompressedRistrettoPublic::from(RistrettoPublic::from_random(rng));
```
eranrund, line 186:
```suggestion
    let ingress_key = CompressedRistrettoPublic::from_random(rng);
```

fog/view/server/src/server.rs
```diff
@@ -397,21 +403,31 @@ where
...
-        let highest_processed_block_count = self
-            .enclave_block_tracker
-            .highest_fully_processed_block_count(&ingestable_ranges, &missing_block_ranges);
-        shared_state.highest_processed_block_count = highest_processed_block_count;
+        if shared_state.highest_processed_block_count != highest_processed_block_count {
+            shared_state.highest_processed_block_count = highest_processed_block_count;
+            self.last_unblocked_at = Instant::now();
+        } else if self.last_unblocked_at.elapsed() >= Duration::from_secs(60) {
+            if let Some(reason_we_stopped) = reason_we_stopped {
+                log::warn!(self.logger, "We seem to be stuck at highest_processed_block_count = {} for at least a minute... we are blocked on an ingress key making progress: {:?}", highest_processed_block_count, reason_we_stopped);
```
eranrund, line 424: I suspect this will be noisy since our network is not very active right now. We probably want some kind of Prometheus alerting if highest processed block count is too far from the block count reported by nodes. Looping in @jgreat @joekottke in case they want to track this.

fog/view/server/tests/smoke_tests.rs
```diff
@@ -8,8 +8,12 @@
 // It exercises both the ingest enclave, and the fog-related crypto that makes
 // its way into the client.
 
+// FIXME: Fix unreachable code
+#![allow(unreachable_code)]
```
eranrund, line 12: Reminder that this is here.

fog/view/server/src/block_tracker.rs
```diff
@@ -1,85 +1,104 @@
 // Copyright (c) 2018-2021 The MobileCoin Foundation
 
-use fog_recovery_db_iface::{IngestInvocationId, IngestableRange};
-use fog_types::common::BlockRange;
+use fog_recovery_db_iface::IngressPublicKeyRecord;
 use mc_common::logger::{log, Logger};
+use mc_crypto_keys::CompressedRistrettoPublic;
 use std::collections::HashMap;
 
 /// A utility object that keeps track of which block number was processed for every known ingest
```
eranrund, line 8: `for every known ingres key`.

fog/view/server/src/block_tracker.rs
```diff
@@ -90,61 +109,49 @@ impl BlockTracker {
                 next_block_count,
             );
 
-            // If the block has been reported as missing, we can advance since we don't need to do
-            // any processing for it.
-            if Self::is_missing_block(missing_block_ranges, next_block_index) {
+            // If the next block index we are checking doesn't exist yet, then we definitely
+            // can't advance the highest processed block count.
+            // This breaks the loop if ingress_keys set is empty.
+            if highest_known_block_count < next_block_count {
```
eranrund, line 115: I believe `highest_known_block_count` does not advance when there are missing ranges/gaps. What happens if there's a gap, e.g. key 1 processed blocks 0-100 and then crashed. the network then did 100 blocks with no key being active. key 2 only started at block 200 because it took us time to restart ingest for whatever reason. wouldn't this cause the view server to never advance past 100 since no key provides 101?

fog/view/server/src/block_tracker.rs
```diff
@@ -461,104 +434,30 @@ mod tests {
...
-            block_tracker.highest_fully_processed_block_count(&[], &[BlockRange::new(0, 10)]),
-            10
+            block_tracker.highest_fully_processed_block_count(&[]),
+            (0, None)
         );
     }
 
-    // Multiple missing ranges are handled appropriately
-    #[test_with_logger]
-    fn highest_fully_processed_block_missing_blocks_consecutivie_missing1(logger: Logger) {
```
eranrund, line 494: This test relates to my comment above with how the view server advances to the next block. I think if the ingress keys are not all consecutive and somehow there is a gap then the view server will stop progressing (unless I misunderstood something).

fog/view/server/src/block_tracker.rs
```diff
@@ -568,233 +467,221 @@ mod tests {
...
+            block_tracker.highest_fully_processed_block_count(&[rec1.clone(), rec2.clone()]),
+            expected,
+        );
+    }
+
+    // A block tracker with a multiple ingestable ranges waits for both of them,
+    // but makes progress when the slow one is reported lost
+    #[test_with_logger]
+    #[ignore]
+    fn highest_fully_processed_block_tracks_multiple_recs_some_lost1(_logger: Logger) {
```
eranrund, line 680: Good test case, commenting to increase the chances of not forgetting to implement it :)

fog/view/server/src/db_fetcher.rs
```diff
@@ -268,110 +229,84 @@ impl<DB: RecoveryDb + Clone + Send + Sync + 'static> DbFetcherThread<DB> {
...
-            let shared_state = self.shared_state();
-            (
-                shared_state.ingestable_ranges.clone(),
-                shared_state.missing_block_ranges.clone(),
-            )
-        };
+        let ingress_keys = self.shared_state().ingress_keys.clone();
 
-        let next_block_index_per_invocation_id = self.block_tracker.next_blocks(&ingestable_ranges);
+        let next_block_index_per_ingress_key = self.block_tracker.next_blocks(&ingress_keys);
```
eranrund, line 234: There's something suboptimal here: `next_blocks()` is aware of the last scanned block, but it doesn't do anything with that information. This means the code in this function will repeatedly try to load block data that we actually know is not yet present. I don't think it's a big deal right now but this should probably be documented as potential optimization.

eranrund (commented):
I think overall this looks excellent. My biggest concern after reading this is not having enough confidence that some odd cases behave properly, specifically:
1) How things behave when there are overlapping ingests with separate keys. I think this should never happen because that would also mess up the report? However nothing is going to enforce that as far as I can tell. 
2) How would things behave if there is a period of time during which no ingest keys are scanning.

fog/sql_recovery_db/src/lib.rs
garbageslam: thank you, done

fog/view/server/src/server.rs
garbageslam: maybe we should make a grpc API to get this info? like, highest processed block count and reason we stopped?

then we can get it when we need it and don't spam the logs?

garbageslam: we could also make the number of seconds a config parameter, so it can be adjusted or turned off

eranrund: A grpc api endpoint for debugging sounds good. If we're okay making it user-facing then that should be an easy change. I can't think of an obvious reason why not.

fog/view/server/src/server.rs
garbageslam: so here's the issue i see there:

1) We can alert based on highest_processed_block_count being behind the block chain, by making a prometheus counter
2) The piece of information you need then is, the hex of the ingress key that we are stuck on
3) Then we go look at the ingest servers and see if they are all dead or one of them just needs to be activated
4) If they are all dead then we report a lost key to the database (literally that hex string)

So you really kind of want this "reason_we_stopped" value
I don't see how I can stick that value in a prometheus metric, or maybe I just don't know enough about prometheus? it seems like a logs / kibana thing

my other idea was, we could make a grpc API so that the view server can just be asked what key it is stuck on, and that's how you get the info rather than the logs?

lmk what you think

garbageslam: i have demoted this to debug for now, i'm going to make a separate JIRA ticket to expose this somehow

garbageslam:
squashed and rebased 28 commits (!)

fog/view/server/src/block_tracker.rs
garbageslam: i think i see what you are saying

in an earlier version, i was using rec.last_scanned_block to decide the highest_known_block_count, then i cut that because there were comments in other tests that that might cause problems and i wasn't sure it was needed

but i think if we don't use that at least for highest_processed_block_count, it's wrong for the reason you say,
and probably it should use that in next_blocks, for the optimization that you mentioned in a different comment

going to think about this, and start by writing tests that would trigger the bad scenario and confirm that it is an issue. then make this change and confirm that it fixes the test if that is indeed the case.

garbageslam:
this is now green after rebase, and i have kicked a deploy. i'm going to go back and iterate on unit tests now

Cargo.lock
```diff
@@ -2609,7 +2622,7 @@ dependencies = [
  "cargo-emit 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "crc 1.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "curve25519-dalek 3.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "displaydoc 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "displaydoc 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
```
jcape, line 2625: This seems to be happening with all the crates in the mobilecoin submodule, even though this causes builds to (properly) fail when a displaydoc error message would be truncated...

fog/ingest/client/src/config.rs
```diff
@@ -23,6 +25,13 @@ fn parse_duration_in_seconds(src: &str) -> Result<Duration, std::num::ParseIntEr
     Ok(Duration::from_secs(u64::from_str(src)?))
 }
 
+fn parse_ristretto_hex(src: &str) -> Result<CompressedRistrettoPublic, String> {
+    let mut key_bytes = [0u8; 32];
+    hex2bin(src.as_bytes(), &mut key_bytes[..])
+        .map_err(|err| format!("Hex decode error: {:?}", err))?;
+    Ok(CompressedRistrettoPublic::from(&key_bytes))
```
jcape, line 32: At minimum, this should use [`hex::decode_to_slice()`](https://docs.rs/hex/0.4.3/hex/fn.decode_to_slice.html), which is already a dep here, instead of adding `hex2bin`.

Best-case scenario is that CompressedRistrettoPublic simply implements [`hex::FromHex`](https://docs.rs/hex/0.4.3/hex/trait.FromHex.html).

jcape (commented):
I'm not sure I'm in a good place to give detailed commentary on the actual work here, but I did notice some low-hanging fruit, namely the displaydoc 1.7 -> 2.0 change seems to have been reverted, and the `hex2bin` usage is unnecessary, since `hex` does the same thing (and can be build in non-std, AFAIK).

Cargo.lock
garbageslam: thank you, i will investigate

garbageslam:
this failed in deploy like this:

activate-fog-ingest:

```
+ /usr/local/bin/fog_ingest_client --uri insecure-fog-ingest://fog-ingest1.diogenes.svc.cluster.local:3226 --retry-seconds 30 activate
2021-03-25 03:01:34.236118362 UTC WARN Creating insecure gRPC connection to fog-ingest1.diogenes.svc.cluster.local:3226, mc.fog.cxn: insecure-fog-ingest://fog-ingest1.diogenes.svc.cluster.local:3226/, mc.module: mc_util_grpc::grpcio_extensions, mc.src: mobilecoin/util/grpc/src/grpcio_extensions.rs:46
thread 'main' panicked at 'rpc failed: Operation { error: Grpc(RpcFailure(RpcStatus { status: 9-FAILED_PRECONDITION, details: Some("activate: Peer backup error: Failed to set peer insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/ peers list") })), total_delay: 30s, tries: 301 }', fog/ingest/client/src/main.rs:96:43
```


fog-ingest1:

```
2021-03-25 02:45:37.132532252 UTC ERRO Tried to send our peer list to peer insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/, but despite successful status the peer list is still wrong! Expected: [insecure-igp://fog-ingest1.diogenes.svc.cluster.local:8090/, insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/], Found: [insecure-igp://fog-ingest1.diogenes.svc.cluster.local:8090/, insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/], mc.local_node_id: fog-ingest1.diogenes.svc.cluster.local:8090, mc.app: fog_ingest_server, mc.local_node_id: fog-ingest1.diogenes.svc.cluster.local:8090, mc.module: fog_ingest_server::controller, mc.src: fog/ingest/server/src/controller.rs:1305
2021-03-25 02:45:37.132715153 UTC ERRO activate: Peer backup error: Failed to set peer insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/ peers list, rpc_request_id: 111, rpc_client_id: 9044d643dc4fe0ab9e1151a4e4398f334060789e0c8472cb6624842b3d297bbb, mc.local_node_id: fog-ingest1.diogenes.svc.cluster.local:8090, mc.app: fog_ingest_server, mc.local_node_id: fog-ingest1.diogenes.svc.cluster.local:8090, mc.module: mc_util_grpc, mc.src: mobilecoin/util/grpc/src/lib.rs:153
```

fog-ingest2:

```
2021-03-25 03:09:48.841313215 UTC WARN The new set of peers did not contain a URI with our responder-id. We added our URI to the set: [insecure-igp://fog-ingest1.diogenes.svc.cluster.local:8090/, insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/] <-- insecure-igp://fog-ingest2.diogenes.svc.cluster.local:8090/, mc.local_node_id: fog-ingest2.diogenes.svc.cluster.local:8090, mc.app: fog_ingest_server, mc.local_node_id: fog-ingest2.diogenes.svc.cluster.local:8090, mc.module: fog_ingest_server::controller, mc.src: fog/ingest/server/src/controller.rs:823
```

garbageslam:
I have tried to make it so that it doesn't now test for equality of URIs, but tests for equality of responder ids. So hopefully that is simpler, and the debugging output should be better

Cargo.lock
garbageslam: I think i messed up the submodule when i rebased

garbageslam:
i think what happened is i messed up the submodule when i rebased

fog/ingest/client/src/config.rs
garbageslam: using hex::decode_to_slice now

garbageslam:
this passed deploy

fog/view/server/src/server.rs
garbageslam: this is now fog-425

fog/view/server/src/db_fetcher.rs
garbageslam: :+1: this is fog-426

garbageslam:
@eranrund thank you for detailed review comments!

I think I have address these concerns with four new unit tests, in most recent commit:

> My biggest concern after reading this is not having enough confidence that some odd cases behave properly, specifically:
>
> 1. How things behave when there are overlapping ingests with separate keys. I think this should never happen because that would also mess up the report? However nothing is going to enforce that as far as I can tell.
> 2. How would things behave if there is a period of time during which no ingest keys are scanning.

and

> I believe highest_known_block_count does not advance when there are missing ranges/gaps. What happens if there's a gap, e.g. key 1 processed blocks 0-100 and then crashed. the network then did 100 blocks with no key being active. key 2 only started at block 200 because it took us time to restart ingest for whatever reason. wouldn't this cause the view server to never advance past 100 since no key provides 101?

and

> This test relates to my comment above with how the view server advances to the next block. I think if the ingress keys are not all consecutive and somehow there is a gap then the view server will stop progressing (unless I misunderstood something).

I believe the test `highest_fully_processed_block_tracks_retired_key_followed_by_gap`
and `highest_fully_processed_block_tracks_retired_key_concurrent_with_active_both_lost`
both speak to what happens in point (2) and these last two comments.

I think that, your analysis is right that, if there is a gap, highest_known_block_index won't cross the gap, and can only get larger if new blocks are reported as processed after the gap.

However, the thing that controls if new blocks get processed is the `next_blocks` function, and `next_blocks` is not limited by hihgest processed block count and doesn't care about the limit. As long as there is a new key with a start_block after the gap, `next_blocks` will say we should load the start block for that key. And once we do load that block and report it processed, `highest_known_block_count` will get larger, and then highest_processed_block_count can get larger, and so cross the gap. This is what happens in the tests right now.

I agree that it might be slightly better if we use the max of `last_scanned_block` as well when determining where to cap the highest processed block count, because those blocks are also "known", and it makes it easier to see that the thing won't get stuck, which is good. However, if it's not a bug right now, I'd like to do that in a later ticket, because I think this version works as is.

In regards to point (1): I think this is now covered effectively by the tests:

* `highest_fully_processed_block_tracks_multiple_recs`
* `highest_fully_processed_block_tracks_multiple_recs2
* `highest_fully_processed_block_tracks_retired_key_concurrent_with_active`
* and to some extent `highest_fully_processed_block_tracks_retired_key_concurrent_with_active_both_lost`

I think the short answer is, if there are two keys, then either of them can block us. And the calculation of whether either of them is blocking us is made independently. It's not blocking us if the calculation in `covers_block_index` says that key doesn't need to provide the next block, otherwise it is and we have to wait for it.

fog/view/server/src/db_fetcher.rs
garbageslam: i think this behavior is correct -- the reason is, if the new ingress key starts at 75, Bob may have an RNG and some transactions that were scanned by that ingest say at blocks 77, 88, 99.
He may find those transactions by view key scanning, but if he later gets a TxOut in block 124, that will be the 4th output of the RNG. But from point of view of fog view protocol, he is still on the first output of his RNG. So even though those blocks are reported missing (because of a different key), fog view still needs to scan them, or else Bob will get blocked on rng value 4 and not find TxOut in block 124

fog/view/server/src/db_fetcher.rs
garbageslam: i think this is okay -- i think ingest only actually scans consecutive sequences.

if we wanted to change this, we could make the database enforce that consecutive sequences are scanned.

i don't think this is the wrong behavior for the view server to have in this scenario

garbageslam: maybe its bad because, if there are gaps like this, reporting the key as lost won't fix it?

garbageslam: idk, i can think of a few ways to try to fix this in the database if it actually happened, but i don't think it will actually happen because i think ingest asserts that it always processes consecutive blocks

sugargoat (approved):

fog/test_infra/src/db_tests.rs
garbageslam: done

fog/view/server/src/block_tracker.rs
```diff
@@ -464,344 +515,790 @@ mod tests {
...
-                ingestable_range.start_block + i,
+                block_tracker.next_blocks(&[rec2.clone()]),
+                expected_state,
+                "i = {}",
+                i
             );
 
-            block_tracker.block_processed(ingestable_range.id, ingestable_range.start_block + i);
+            // Make the block "exist" and "load" it
+            rec1.last_scanned_block = Some(index);
```
eranrund: HMMMM why is this changing `rec1`? Doesn't seem like `rec1` is used so this is probably some leftover

fog/view/server/src/block_tracker.rs
```diff
@@ -464,344 +515,790 @@ mod tests {
...
 
+            // Check that highest fully processed block count matches what we expect
+            let expected = (index + 1, None);
             assert_eq!(
-                block_tracker.highest_fully_processed_block_count(
-                    &[ingestable_range.clone()],
-                    &missing_ranges
-                ),
-                ingestable_range.start_block + i + 1,
+                block_tracker.highest_fully_processed_block_count(&[rec2.clone()]),
```
eranrund: Why isn't `rec1` being passed into this call? I believe the view server will always pass all the records...

fog/view/server/src/db_fetcher.rs
```diff
@@ -503,95 +468,146 @@ mod tests {
...
+                    pubkey_expiry: 0,
+                    retired: false,
+                    lost: false,
+                },
+                last_scanned_block: Some(29),
+            }]
+        );
+
         // Add more blocks but this time leave a hole between the previous blocks and
         // the new ones. They should not get picked up untill a missed blocks
```
eranrund, line 496:
```suggestion
        // the new ones. They should not get picked up until a missed blocks
```

fog/view/server/src/db_fetcher.rs
```diff
@@ -503,95 +468,146 @@ mod tests {
...
+            ingress_keys,
+            vec![IngressPublicKeyRecord {
+                key: ingress_key.clone(),
+                status: IngressPublicKeyStatus {
+                    start_block: 10,
+                    pubkey_expiry: 0,
+                    retired: false,
+                    lost: false,
+                },
+                last_scanned_block: Some(49),
```
eranrund, line 523: This was initially confusing to me, but the 49 value comes from MAX(block_index) on the processed blocks db table, which makes sense since we did add blocks [40-50) above. I would document this for future us.

fog/view/server/src/db_fetcher.rs
```diff
@@ -503,95 +468,146 @@ mod tests {
...
+        );
 
         sleep(Duration::from_secs(1)); // Supposedly enough time for at least some blocks to get picked up.
 
-        assert!(!db_fetcher.shared_state().fetched_records.is_empty());
+        assert!(db_fetcher.shared_state().fetched_records.is_empty());
+
+        // Retire our key at block 45, and provide blocks 30-39 (we previously provided
+        // 40-49)
+        // We should only get block data for blocks 30-34.
```
eranrund, line 533: Where's the 34 coming from?

eranrund (changes requested):
Thank you for so thoroughly addressing my comments regarding testing. Aside from one discovery in one of the tests this looks good to me.

fog/view/server/src/db_fetcher.rs
garbageslam: done

fog/view/server/src/db_fetcher.rs
garbageslam: i think its a typo, i think its actually 44 , and then its because thats the last block before expiry at 45. but i'm still confirming that

eranrund: I reached a similar conclusion.

garbageslam: i have updated the comment, i think it was a typo

garbageslam:
thank you for detailed review! i think those are addressed now, please let me know if you see anything else

garbageslam:
i will try re-reading the PR again also

eranrund (approved):
🙏

PR #19, opened 2021-03-16
fog conformance tests: add a happy path 2nd server + retire test

eranrund:
Soundtrack of this PR: https://soundcloud.com/janoma/janoma-live-sisyphos-the-bronny-bounce

### Motivation

Currently conformance tests only utilize a single ingest server. We want to make the tests more elaborate and have them cover multiple scenarios that involve numerous ingest servers. This PR is the first step in that direction.

### In this PR
* Change fog ingest client utility to output JSON ingest summary data and consume it in the Python code
* Uprev `mobilecoin`
* Add a test that spawns a 2nd ingest server, instructs the 1st one to retire, send transactions using the second server's ingest key and verify they are found using the test client.

### Future Work
* More elaborate tests.

### DIFF
--- a/fog/ingest/client/src/main.rs
+++ b/fog/ingest/client/src/main.rs
@@ -2,12 +2,14 @@
 
 //! Fog Ingest client
 
+use fog_api::ingest_common::IngestSummary;
 use fog_ingest_client::{
     config::{IngestConfig, IngestConfigCommand},
     ClientResult, FogIngestGrpcClient,
 };
 use fog_uri::FogIngestUri;
 use mc_common::logger::{create_root_logger, log, Logger};
+use serde_json::json;
 use std::{str::FromStr, sync::Arc};
 use structopt::StructOpt;
 
@@ -54,12 +56,14 @@ fn main() -> ClientResult<()> {
 fn get_status(logger: &Logger, ingest_client: &FogIngestGrpcClient) -> ClientResult<()> {
     let status = ingest_client.get_status().expect("rpc failed");
     log::info!(logger, "Status: {:?}", status);
+    println!("{}", ingest_summary_to_json(&status));
     Ok(())
 }
 
 fn new_keys(logger: &Logger, ingest_client: &FogIngestGrpcClient) -> ClientResult<()> {
     let status = ingest_client.new_keys().expect("rpc failed");
     log::info!(logger, "Done, status: {:?}", status);
+    println!("{}", ingest_summary_to_json(&status));
     Ok(())
 }
 
@@ -72,6 +76,7 @@ fn set_pubkey_expiry_window(
         .set_pubkey_expiry_window(pubkey_expiry_window)
         .expect("rpc failed");
     log::info!(logger, "Done, status: {:?}", status);
+    println!("{}", ingest_summary_to_json(&status));
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


tools/fog-local-network/local_fog.py
```diff
@@ -80,7 +81,7 @@ def start(self):
             f'--peer-listen-uri=insecure-igp://localhost:{self.peer_port}/',
             f'--ias-api-key={IAS_API_KEY}',
             f'--ias-spid={IAS_SPID}',
-            f'--local-node-id localhost:{self.client_port}',
```
garbageslam, line 83: thank you!

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -564,23 +557,16 @@ def run(self):
...
-        self.follow_up_balance_check(wallet0_from4a, {5: 10}, [5])
-        self.follow_up_balance_check(wallet1_from4a, {5: 6}, [5])
-        self.follow_up_balance_check(wallet2_from4a, {5: 0}, [5])
-        self.follow_up_balance_check(wallet3_from4a, {5: 0}, [5])
-        self.follow_up_balance_check(wallet4_from4a, {5: 0}, [5])
+        for (wallet0, wallet1, wallet2, wallet3, wallet4) in (
+            (wallet0_from1, wallet1_from1, wallet2_from1, wallet3_from1, wallet4_from1),
+            (wallet0_from4, wallet1_from4, wallet2_from4, wallet3_from4, wallet4_from4),
+            (wallet0_from4a, wallet1_from4a, wallet2_from4a, wallet3_from4a, wallet4_from4a),
+        ):
```
garbageslam, line 564: interesting -- i did not know this works in python

one thing is, when the test fails, you won't get a line number showing which one it is, but it there is a "name" for the wallet like wallet3_from1 that gets reported when balance check reports an error

eranrund: I'll look into ensuring the name associated with the wallets is printed before each check.

garbageslam (approved):
this looks good to me, thank you!

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -784,77 +720,130 @@ def run(self):
...
+        # Both ingests should currently be active
+        status = self.fog_ingest.get_status()
+        assert status["mode"] == "Active", status
+
+        status = self.fog_ingest2.get_status()
+        assert status["mode"] == "Active", status
+
+        # Add block 9 to ingest and ledger. This should cause ingest1 to become Idle.
+        # Give 1 to everyone
+        # Take 2 from ???
```
mfaulk, line 804: ???

fog/ingest/client/src/main.rs
```diff
@@ -54,12 +56,14 @@ fn main() -> ClientResult<()> {
 fn get_status(logger: &Logger, ingest_client: &FogIngestGrpcClient) -> ClientResult<()> {
     let status = ingest_client.get_status().expect("rpc failed");
     log::info!(logger, "Status: {:?}", status);
+    println!("{}", ingest_summary_to_json(&status));
```
sugargoat, line 59: What's the reason for the printlns? Is this read by the conformance test?

eranrund: Yup, this allows scripts to pick up the output and parse it.

sugargoat: I see - neat :)

sugargoat (approved):

PR #21, opened 2021-03-17
Optional Fog info in public address

voloshyn:
### Motivation

Add ability to create public addresses without Fog


sugargoat (approved):

UkoeHB:
Hi, not sure if you considered this, but if a wallet _doesn't_ support Fog, then it should reject attempts to send money to Fog-enabled addresses (sort of related).

voloshyn:
> Hi, not sure if you considered this, but if a wallet _doesn't_ support Fog, then it should reject attempts to send money to Fog-enabled addresses (sort of related).

Hi, sending from non-fog to fog accounts is a supported use case in MobileCoin's desktop wallet. However, the android-bindings are designed to be used with Fog, so, unlike PublicAddress, an AccountKey cannot be created without fog info using the bindings APIs.

tsegaran (approved):

PR #22, opened 2021-03-17
FOG-408: Make ingest peer uri filtering based on responder id

garbageslam:
This fixes a bug that exists today in fog ingest deployments.
I thought this was only a minor issue because, it would wrongly
add the peer listen uri to the list of peer uris, but then filter
it out and not connect to it. But because the active server replicates
its peer list onto the backups continuously, they will all end up
with this uri in their list, and there is no configuration that will
fix this. Additionally, a server won't be able to filter out it's
own (real) URI that the peers have in their lists, because it doesn't
filter by responder id before this commit.


sugargoat:
Was there a successful deployment with this revision?

garbageslam:
not yet

eranrund (commented):
LGTM once everything is green and happy. Let me know if you want to offload getting it there to me.

sugargoat (approved):
LGTM once successful deploy

jcape (approved):
It's fine as-is, but it really seems like the "is the responder ID valid" test should happen when creating an `IngestPeerUri`, no? Is there ever a case where an `IngestPeerUri` without a responder ID is valid?

garbageslam:
no, i did not create the responder id API -- i made a ticket already to track making responder id infallible

fog/ingest/server/src/config.rs
```diff
@@ -38,7 +38,7 @@ pub struct IngestConfig {
     pub peer_listen_uri: IngestPeerUri,
 
     /// List of all peers in this cluster
-    #[structopt(long)]
+    #[structopt(long, use_delimiter = true)]
```
eranrund, line 41: Why was this needed?

garbageslam: if you want to have it work like `--peers igp://foo,igp://bar,igp://baz,igp://qaz` you need to put `use_delimter = true`. Otherwise, when structopt sees a Vec, you must make it like `--peers igp://foo --peers igp://baz --peers igp://qaz` and that just seems harder to read

garbageslam: here's the issue in structopt repo: https://github.com/TeXitoi/structopt/issues/94

PR #25, opened 2021-03-17
Add ledger clone utility method, make prints more verbose, add a test for restarting the view server

eranrund:
This PR does two things:
1. Adds a utility method to clone an existing ledger/watcher db, this will get used in future tests.
2. Adds a test that restarts the view server and makes sure the new view server loads all data properly.


sugargoat (approved):

jcape (approved):

garbageslam (approved):
thank you!

PR #28, opened 2021-03-22
DRY/clean up parts of fog conformance tests

eranrund:
### Motivation

The Fog conformance test script was getting longer and contained a bunch of repeated code to deal with the over-increasing list of clients we are testing with. Each step would start a new balance check client to perform fresh balance checks while also performing balance checks with all the previously-started clients. The list was maintained by hand which is annoying and unnecessary.

### In this PR
* Create a new `MutliBalanceChecker` object that handles the collection of clients and is capable of performing a balance check on all of them.
* Some cleanups of the "credits" part of the test to make it more readable - each account is on its own lines instead of all accounts mishmashed in one long line.

### Future Work
* Figure out a way to make the script use pytest or some other testing framework in order to make it easier to write multiple tests that are independent of each-other (as opposed to the current single test that is an assortment of steps that build on top of each other).

### DIFF
--- a/tools/fog-local-network/fog_conformance_tests.py
+++ b/tools/fog-local-network/fog_conformance_tests.py
@@ -165,6 +165,7 @@ def parse_balance_check_output(line):
     return result
 
 class BalanceCheckProgram:
+    """An object for starting and controlling the execution of an external balance check program."""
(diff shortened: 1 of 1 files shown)


sugargoat (approved):

garbageslam (approved):
very nice!

that cleans up a lot of boilerplate

PR #29, opened 2021-03-22
WIP: Fog 180 update ingest report cache refresh rate

queenbdubs:
DRAFT of test added for FOG-180 to test get_last_report_published_block_index()


fog/ingest/server/src/controller_state.rs
```diff
@@ -268,6 +273,20 @@ impl IngestControllerState {
...
+        self.last_report_published_block_index
+    }
+
+    /// Set the last block that an ingest report was published for
+    pub fn set_last_report_published_block_index(&mut self, new_val: u64) {
+        self.last_report_published_block_index = new_val;
+    }
+
+    /// Check is the given block index the last block that an ingest report was published for
+    pub fn is_last_report_published_block_index(&self, block_index: u64) -> bool {
```
sugargoat, line 287: Maybe a name like `has_report_expired` would be slightly more clear

fog/ingest/server/src/ingest_service.rs
```diff
@@ -180,6 +180,25 @@ where
...
+    /// The last block that an ingest report was published for
+    pub fn get_last_report_published_block_index_impl(
+        &mut self
+    ) -> Result<GetLastReportPublishedBlockIndexResponse, RpcStatus> {
+        // TODO: The Err(()) => 0 is problematic?
+        let last_report_published_block_index = match self
+            .controller
+            .get_last_report_published_block_index() {
+                Ok(last_report_published_block_index) => last_report_published_block_index,
+                Err(()) => 0
```
sugargoat, line 193:
```suggestion
                Err(e) => return Err(rpc_internal_error("get_last_report_published_block_index", e, &self.logger))
```

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -764,6 +765,54 @@ def run(self):
...
+                     {'account': 3, 'amount': 0},
+                     {'account': 4, 'amount': 0}]
+        key_images7x = [block7_key_images[0]]
+        print("===> Key images spent in Block 7x: ", key_images7x)
+        block7x_key_images = ledger1.add_block(credits7x, key_images7x, fog_pubkey)
+        ledger2.add_block(credits7x, key_images7x, fog_pubkey)
+        print("===> Key images for new transactions in Block 7x: ", block7x_key_images)
+        time.sleep(1)
+
+        # Expecting index=8
```
sugargoat, line 794: Maybe add a assert that current expiry window is 1

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -764,6 +765,54 @@ def run(self):
...
+        index2 = self.fog_ingest.get_last_report_published_block_index()
+        print("===> last report published block index: ", index2)
+
+        assert (index2 != index1) and (index2 > index1)
+
```
sugargoat, line 799: Maybe add a step that changes the expiry window to 3, adds 3 blocks, and verifies after each block that the report hasn't been published until it reaches the third block.

@eranrund - can you check our logic here: the intent is for this test to verify that changing the expiry window is properly handled in conjunction with report publishing (particularly when the report is no longer published on every block).

@queenbdubs explained why reducing the pubkey expiry window is problematic - but we do still need to make sure that the code handles this correctly - both on increase and decrease of expiry window.

IIUC, in the wild, we may increase the expiry window for some reason (maybe if transaction volume is high and blocks are being written quickly?), and if we decreased it later because transaction volume settled down, we would still keep the report valid up until the last block in the expiry window, then decrease it after that block.

fog/ingest/server/src/controller.rs
```diff
@@ -975,6 +978,11 @@ where
         self.recovery_db.get_missed_block_ranges()
     }
 
+    pub fn get_last_report_published_block_index(&self) -> Result<u64, ()> {
+        // Should return be Result<u64, ()> ?
```
eranrund, line 982: Should just return `u64` if it can't error.

fog/ingest/server/src/controller_state.rs
```diff
@@ -268,6 +273,20 @@ impl IngestControllerState {
...
+        return self.last_report_published_block_index == block_index.saturating_sub(self.pubkey_expiry_window)
```
eranrund, line 288: What is this trying to check for?

tools/fog-local-network/fog_conformance_tests.py
eranrund: Yes, this seems correct - I believe increasing should work, decreasing should result in an error.

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -764,6 +765,54 @@ def run(self):
...
+        block7y_key_images = ledger1.add_block(credits7y, key_images7y, fog_pubkey)
+        ledger2.add_block(credits7y, key_images7y, fog_pubkey)
+        print("===> Key images for new transactions in Block 7y: ", block7y_key_images)
+        time.sleep(3)
+
+        # Expecting index=9 (works when sleep>1, reported index=8 when sleep=1)
+        index3 = self.fog_ingest.get_last_report_published_block_index()
+        print("===> last report published block index: ", index3)
+
+        assert (index3 != index2) and (index3 > index2)
```
eranrund, line 814: I think the assert should check for exact values since in this test we know exactly at which block we expect to be.

eranrund (commented):
Nice progress! I think keeping track of the last report published in the state is a good move. I also think there is no need to add another RPC call to return it since it is already included in the GetStatus call. I am confused about what the new check is trying to achieve? Not having the report go backwards?

fog/ingest/server/src/controller.rs
```diff
@@ -336,10 +336,13 @@ where
             // This also returns an ingress key state which indicates if publishing was
             // successful, it may fail if the key is retired. Then we can check
             // if we are still needed to be active.
-            if state.is_active() {
+            if state.is_active() && state.is_last_report_published_block_index(block.index) {
```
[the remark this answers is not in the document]
queenbdubs: Sarah mentioned this too...We thought maybe `has_report_expired()` is a better name?  Maybe `is_report_expired()`?   To answer your question further down, this function is supposed to check to see if the current block.index has reached the pubkey expiry window.  If it has, then when we successfully publish the report and update the variable to store the block index of the last report published.

When looking further into it though, I think the logic for the `has_report_expired()` function is incorrect.  I think we actually need the absolute value of  `pubkey_expiry_window - block.index`, so that would mean the function should be this:
(I didnt find a Rust abs() function on u64 so this is implementing that)
```
pub fn has_expired_report(&self, block_index: u64) -> bool {
        let compare : u64;
        if block_index < self.pubkey_expiry_window {
            compare = self.pubkey_expiry_window.saturating_sub(block_index);
        } else {
            compare = block_index.saturating_sub(self.pubkey_expiry_window);
        }
        self.last_report_published_block_index == compare
    }
```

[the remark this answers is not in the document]
eranrund: I think there is a potentially more elegant solution. Since we now keep track of what is the last pubkey expiry we published, what would happen if `publish_report` would change to:
1) not publish a new report if the calculated pubkey expiry is lower than the last published one
2) be in charge of updating `state.last_report_published_block_index` if a new report has been successfully published
It seems to be that this will both simplify the logic, and remove the need to do separate book-keeping using `set_last_report_published_block_index` since I don't see when we would want to call that without previously calling `publish_report`.

fog/api/proto/ingest.proto
```diff
@@ -60,6 +60,8 @@ service AccountIngestAPI {
 
     /// Get list of missed block ranges.
     rpc GetMissedBlockRanges(google.protobuf.Empty) returns (GetMissedBlockRangesResponse) {}
+
+    rpc GetLastReportPublishedBlockIndex (google.protobuf.Empty) returns (GetLastReportPublishedBlockIndexResponse) {}
```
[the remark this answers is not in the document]
queenbdubs: I'm actually not sure!  If that's a better way to go, happy to investigate as I get more familiar with the codebase.  What's the starting point for where I would begin to add that?

fog/ingest/server/src/controller.rs
queenbdubs: Sarah suggested an rpc_internal_error down in the ingest_service.rs file changes below...I think that eventually bubbles up to here?

fog/api/proto/ingest.proto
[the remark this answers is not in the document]
eranrund: Take a look at the `GetStatus` call. It returns an `IngestSummary` object - and it looks like `last_report_published_block_index` was already added to it.

fog/ingest/server/src/controller.rs
eranrund: `rpc_internal_error` is in the GRPC service implementation, which is what calls this method. However, since this is a simple getter, there is no scenario it which it can fail.

.gitignore
```diff
@@ -69,3 +69,6 @@ lock.mdb
 
 # Homebrew bundle
 /Brewfile.lock.json
+
+# pycache
+**/__pycache__
```
jcape, line 74: This doesn't appear to be working, `tools/fog-local-network/__pycache__/local_fog.cpython-36.pyc` is part of the PR.

```suggestion
__pycache__
```

fog/ingest/server/src/controller.rs
queenbdubs: uneeded since using get-status now

fog/ingest/server/src/controller_state.rs
queenbdubs: uneeded since using get-status now

fog/ingest/server/src/ingest_service.rs
queenbdubs: unneeded since using get-status now

fog/ingest/server/src/controller.rs
```diff
@@ -691,28 +694,37 @@ where
...
-        // tombstone block. But scanning the tombstone block as well
-        // may help deal with off-by-one errors somewhere else, and doesn't really hurt.
-        if key_status.retired && start_block > key_status.pubkey_expiry {
-            log::warn!(self.logger, "When activating, we found out our key has already been retired and there is no remaining work to do. Activation is canceled: our start_block = {}, key_status = {:?}", start_block, key_status);
-            return Err(Error::KeyAlreadyRetired(our_pubkey));
-        }
+        match self.publish_report(&our_pubkey, &mut state) {
+            Ok(None) => {
+                // TODO: is this ok to return?
+                return Ok(IngestSummary::new());
```
eranrund, line 700: This should be `return Ok(self.get_ingest_summary());`

.gitignore
queenbdubs: Saw that the __pycache__ exclusion is already listed earlier in the .gitignore

tools/fog-local-network/fog_conformance_tests.py
```diff
@@ -869,35 +1003,95 @@ def run(self, args):
...
+            [{16: 14, 17: 14}, [17]],
+            [{16: 13, 17: 14}, [17]],
+            [{16: 4, 17: 5}, [17]],
+            [{16: 4, 17: 5}, [17]],
+            [{16: 17, 17: 18}, [17]],
         ])
 
+        # TODO: Expecting the server to be Idle here, but it's still Active...
+        status = self.fog_ingest.get_status()
+        print("Ingest server status after block 16: ", status)
```
queenbdubs, line 1062: This is an outstanding issue....After block 16 is published, I was expecting the server to retire and be Idle here, but the status still reports it as Active.  Not sure why or what's missing.

fog/ingest/server/src/controller.rs
```diff
@@ -1027,20 +1042,103 @@ where
...
+        // However, if we would have increased the window, so:
+        //    block_index=7 with window=102 gives a pubkey_expiry=109
+        // ...we do publish, b/c the previous index's pubkey_expiry (106)
+        //    is smaller than the current index's pubkey_expiry (109).
+        //
+        // We don't want to publish on every index, only every N indexes,
+        // where N is the size of the window.
+        // ex.
+        // If block_index=1 with window=100, we publish a report on indexes
+        // 101, 201, 301, 401, etc.
```
queenbdubs, line 1091: The previous logic just checked if the previous pubkey_expiry was smaller than the current pubkey_expiry, but we still don't want to publish a report every single time.  So I amended the if-logic below as an attempt to only publish every N indexes (where N is the size of the window).

eranrund: @garbageslam is that true? Why would we not want to publish a new report every index?
Publishing every window size seems like a meaningful change to me so I am wondering what is the reason behind it.
What happens if we fail to publish at this reduced interval? Seems like we'll need a much more aggressive retry policy. We could also run into an issue where the consensus network is slightly ahead of fog, in which case users attempting to pay fog recipients will have to wait until fog ingest catches up (e.g. if our window is 100 and we're starting at block 50, we will publish at 50, 150, 250, etc. At block 149 a user might grab a report that is only valid for one block and consensus might already be past it, while ingest might still be at for example block 140 chugging along towards 150 but until it gets there users cannot make payments).

eranrund: To add to that, my personal preference is to keep the changeset to a minimum. If we do want to reduce the publishing frequency we can tackle that in a followup PR.

queenbdubs: oh hmm..I thought this ticket was to do just that (reduce publishing frequency).  But only in the case that we increase the window.  Right now, the controller state's set_pubkey_expiry_window() function doesn't let you reduce the window.  So I thought that was the separate ticket if we wanted to handle that case.  Let me see if we can get some clarification.

garbageslam:
I'm going to close this PR because we merged #116 which went in a different direction for this issue (not reducing frequency of publishing report, but only reducing frequency of calling `update_enclave_report_cache`)

Thank you for your work on this!

PR #34, opened 2021-03-31
Uprev MobileCoin Submodule to 61349cbc

sugargoat:


kylefleming (approved):

fog/ledger/enclave/trusted/Cargo.lock
```diff
@@ -1286,6 +1292,9 @@ dependencies = [
 name = "rand_core"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "getrandom 0.1.13 (registry+https://github.com/rust-lang/crates.io-index)",
+]
```
jcape, line 1297: This is sketchy

sugargoat: Oh interesting - that was the result of running `cargo fetch` - any idea what's going on there?

jcape (approved):

fog/ledger/enclave/trusted/Cargo.lock
jcape: `Cargo.lock` is recording feature-unified dependencies (in this case, the build-dependencies of `mc-attest-core` changed to include `rand_hc` and `rand_core` to do `MC_SEED` reasonably), but I don't believe `getrandom` is actually being compiled into the enclaves here.

PR #35, opened 2021-03-31
[SWIFT-164] Add support for slip10 key derivation to LibMobileCoin

kylefleming:
### Motivation

The new MobileCoin mnemonic key derivation spec is based on SLIP-10. This PR adds support for it to LibMobileCoin.

Note: It removes support for the old, deprecated derivation method (using "root entropy"). This is okay because LibMobileCoin is in a beta/prerelease state and isn't yet used in production in an clients.

### In this PR

* Adds support for slip10 key derivation to LibMobileCoin
* Upgrades `mc-android-bindings` from using `bip39` to using `tiny-bip39`.

### DIFF
--- a/libmobilecoin/include/slip10.h
+++ b/libmobilecoin/include/slip10.h
@@ -9,35 +9,25 @@
 extern "C" {
 #endif
 
-/* ==== Types ==== */
-
-typedef struct _McSlip10Indices McSlip10Indices;
-
-/* ==== McSlip10Indices ==== */
-
-McSlip10Indices* MC_NULLABLE mc_slip10_indices_create(void);
-
-void mc_slip10_indices_free(
-  McSlip10Indices* MC_NULLABLE indices
-);
-
-bool mc_slip10_indices_add(
-  McSlip10Indices* MC_NONNULL indices,
-  uint32_t index
-)
-MC_ATTRIBUTE_NONNULL(1);
-
 /* ==== McSlip10 ==== */
 
 /// # Preconditions
 ///
-/// * `out_key` - length must be >= 32.
-bool mc_slip10_derive_ed25519_private_key(
-  const McBuffer* MC_NONNULL seed,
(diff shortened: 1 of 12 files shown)


libmobilecoin/include/slip10.h
```diff
@@ -9,35 +9,25 @@
...
 /* ==== McSlip10 ==== */
 
 /// # Preconditions
 ///
-/// * `out_key` - length must be >= 32.
-bool mc_slip10_derive_ed25519_private_key(
-  const McBuffer* MC_NONNULL seed,
-  const McSlip10Indices* MC_NONNULL path,
-  McMutableBuffer* MC_NONNULL out_key
+/// * `mnemonic` - must be a nul-terminated C string containing valid UTF-8.
```
sugargoat, line 16:
```suggestion
/// * `mnemonic` - must be a null-terminated C string containing valid UTF-8.
```

libmobilecoin/src/slip10.rs
```diff
@@ -1,47 +1,47 @@
...
-pub extern "C" fn mc_slip10_indices_add(indices: FfiMutPtr<McSlip10Indices>, index: u32) -> bool {
-    ffi_boundary(|| {
-        indices.into_mut().push(index);
-    })
-}
-
 /// # Preconditions
 ///
-/// * `out_key` - length must be >= 32.
+/// * `mnemonic` - must be a nul-terminated C string containing valid UTF-8.
```
sugargoat, line 12:
```suggestion
/// * `mnemonic` - must be a null-terminated C string containing valid UTF-8.
```

sugargoat (approved):

voloshyn (approved):
Android bindings LGTM

libmobilecoin/include/slip10.h
kylefleming: I've been using `nul` everywhere. I think if we're going to change it to `null`, we should do it all at once.

PR #38, opened 2021-04-01
GRPC Fog Conformance Tests

eranrund:
### Motivation

We want to run the fog conformance tests against the mobile SDKs. Currently, the fog conformance tests require running a program and communicating with it over `stdin`/`stdout`. This does not play nicely with wallets running on mobile emulators.
The emulators, however, allow us to start a GRPC service - as such, we'd like to be able to have the balance checking mechanism work using GRPC instead of `stdin`/`stdout`.

### In this PR
* Introduce a new API (inside `remote_wallet.proto`) for exposing a wallet over GRPC.
* Introduce a new utility, `sample_paykit_remote_wallet`, which provides a GRPC server implementation of this API using the Rust sample paykit
* Change the fog conformance tests to perform balance checks using the GRPC API, replacing the current child-process-over-stdin/out mechanism.
* Update CI to start the new `sample_paykit_remote_wallet` utility before running the fog conformance tests.

### Future Work
* Provide an implementation of the new remote wallet API in Android and iOS apps, so that the conformance tests could be ran against clients written in Java/Swift and not just the Rust sample client code.

### Notes
This changes how fog performance tests are ran. Previously, you simply run `tools/fog-local-network/fog_conformance_tests.py target/debug/balance_check` and this would take care of the rest. After this PR is merged, the process is slightly more involved - see https://github.com/mobilecoinfoundation/fog/blob/8b145e3f8c9b89e6b55e87fea58d17536e48ae79/tools/fog-local-network/README.md#usage


sugargoat (approved):
Awesome

fog/sample-paykit/proto/remote_wallet.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message FreshBalanceCheckRequest {
+    /// Root entropy for the account the new client will be using.
+    bytes root_entropy = 1;
+
+    /// View server URI
+    string view_server_uri = 2;
+
+    /// Ledger server URI
+    string ledger_server_uri = 3;
```
kylefleming, line 42: This should be unified into a single `fog_uri`.

fog/sample-paykit/proto/remote_wallet.proto
```diff
@@ -0,0 +1,70 @@
...
+}
+
+message BalanceCheckResponse {
+    /// A unique client identifier that is used in followup requests to refer to the client.
+    uint32 client_id = 1;
+
+    /// The account balance.
+    uint64 balance = 2;
```
kylefleming, line 50: Not sure if this is entirely critical, but technically an account could have a balance of `> uint64.max`.

A `uint64` + an additional `uint8` would be required to hold the largest possible balance. This is of course probably very unlikely to ever happen in production though, so it might not be a thing we want to worry about. However, for comparison, the Swift SDK does make it a point to handle account balances `> uint64.max` though.

kylefleming (commented):
Partial review (mainly reviewed the `remove_wallet.proto`). The 2 comments about the entropy and about the fog uri I think are important to address.

fog/sample-paykit/proto/remote_wallet.proto
```diff
@@ -0,0 +1,70 @@
...
+    /// Get debug info for a previously-started client.
+    rpc Debug (DebugRequest) returns (DebugResponse);
+
+    /// Drop all previously-started clients.
+    rpc Reset (google.protobuf.Empty) returns (google.protobuf.Empty);
+}
+
+message FreshBalanceCheckRequest {
+    /// Root entropy for the account the new client will be using.
+    bytes root_entropy = 1;
```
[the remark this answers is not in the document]
eranrund: This was written before this change :) I agree it should change to use bip39_entropy once @jcape's work is merged (since this relies on `keyfile`)

fog/sample-paykit/proto/remote_wallet.proto
eranrund: That complicates the local setup. As a reminder, this script runs against services running locally and not a deployed network. If we wanted it to be a single URI then we would need to run `nginx` or something like that to do the splitting, and I am hoping we can avoid that. 
Are the SDKs not going to support specifying separate URLs? @voloshyn did provide me with a version that supports that on Android.

fog/sample-paykit/proto/remote_wallet.proto
eranrund: Since the idea is to use this with the conformance test script, and the amounts are controlled by it, I think it's safe to not worry about >uint64.max and keep this thing simple.

fog/sample-paykit/proto/remote_wallet.proto
kylefleming: The Swift SDK has already dropped support for separating the Fog services.

Aren't there 4 ledger services though, with 2 of them being attested? Could fog view be added to that setup in some way so that it's 5 services?

And ideally fog report would be on the same url as well, although currently it doesn't look like the conformance tests test that part of the system. (But I imagine we will want to expand the conformance tests over time.)

fog/sample-paykit/proto/remote_wallet.proto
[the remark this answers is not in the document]
kylefleming: In the interim, should we use `view_private_key` and `spend_private_key` instead of `root_entropy`? That would at least mean that it's usable with old clients that still use root_entropy key derivation and new clients that are already using mnemonic key derivation (like the ios sdk) and no longer support root_entropy.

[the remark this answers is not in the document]
eranrund: The issue with that is that the fog conformance test python script (the only thing that currently uses this remote wallet thing) does not have a way to do the derivation. It would need to rely on some external utility to do so. Whats in `master` is also currently using root entropy. Unless you feel strongly about this, my preference would be to leave it as is in this PR and fix it in a follow-up PR once the `keyfile` changes land.

fog/sample-paykit/proto/remote_wallet.proto
eranrund: There are currently two server binaries (view and ledger), each listening on their own port. Its true that report is also a separate server but at the moment it is not used in the tests. Similarly to what I said in the other comment, this is currently how `master` works. I would prefer to keep the changes in this PR minimal and move to the single URL+nginx in a separate PR.

voloshyn (approved):
LGTM

PR #45, opened 2021-04-06
Local fog unified fog URI

eranrund:
### Motivation

Our k8/production networks exposes the assortment of fog servers we have (ledger, view, report) over a single URI endpoint (e.g. `fog://fog.test.mobilecoin.com`). Prior to this PR, the local fog network started by the fog conformance tests did not support doing this, making it incompatible with the mobile SDKs.

### In this PR
* Add `nginx` fronting the fog services when running a local fog network.
* Add a flag to the fog conformance test script for overriding the fog uri handed to the remote wallet service. By default, the script will use `insecure-fog://localhost:8200` (8200 is the new `nginx` port). This is good when everything is running on the same machine, but when the remote wallet is running on, for example, a mobile phone or emulator then it needs to know how to connect back to the local fog network. This parameter enables specifying that.
* Fixing a bunch of regressions on documentation/fog local network script. 

### Future Work
There are two more pending items in the ongoing work to make the fog conformance tests play nicely with the mobile SDKs:
* Move from root entropy to bip39 entropy.
* When the remote wallet concept was introduced, the way fog conformance tests is ran has changed such that it now requires manually starting the remote wallet service. This can be improved for cases when everything is running locally (such as CI, or local development that doesn't involve mobile SDKs). Simplifying test running is desirable, so I would like to follow up with a PR that enables the fog conformance test script to take care of starting the sample paykit remote wallet service.

### DIFF
--- a/fog/sample-paykit/src/bin/sample_paykit_remote_wallet.rs
+++ b/fog/sample-paykit/src/bin/sample_paykit_remote_wallet.rs
@@ -87,10 +87,36 @@ impl RemoteWalletService {
         let consensus_client_uri = ConsensusClientUri::from_str("mc://127.0.0.1")
             .expect("Could not create dummy consensus client uri");
 
+        // Figure out the view/ledger URIs and adapt the scheme to match what the Rust
+        // clients expect.
+        let fog_uri = request.get_fog_uri();
+        let (fog_view_uri, fog_ledger_uri) = if fog_uri.starts_with("fog://") {
+            (
+                fog_uri.replace("fog://", "fog-view://").to_string(),
+                fog_uri.replace("fog://", "fog-ledger://").to_string(),
+            )
+        } else if fog_uri.starts_with("insecure-fog://") {
+            (
+                fog_uri
+                    .replace("insecure-fog://", "insecure-fog-view://")
+                    .to_string(),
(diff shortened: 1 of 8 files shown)


fog/sample-paykit/src/bin/sample_paykit_remote_wallet.rs
```diff
@@ -87,10 +87,36 @@ impl RemoteWalletService {
         let consensus_client_uri = ConsensusClientUri::from_str("mc://127.0.0.1")
             .expect("Could not create dummy consensus client uri");
 
+        // Figure out the view/ledger URIs and adapt the scheme to match what the Rust
+        // clients expect.
+        let fog_uri = request.get_fog_uri();
+        let (fog_view_uri, fog_ledger_uri) = if fog_uri.starts_with("fog://") {
+            (
+                fog_uri.replace("fog://", "fog-view://").to_string(),
```
eranrund, line 95: These replaces are needed until the Rust sample paykit switches to using a single URI for all fog services.

sugargoat (approved):

voloshyn (approved):
LGTM

garbageslam:
thank you!

PR #58, opened 2021-04-17
Future clippy

eranrund:
### Motivation

In preparation for a Rust version upgrade I ran into a bunch of small things future clippy will complain about, so here are fixes for them.

### In this PR
* `impl Into` -> `impl From`
* `ORM` -> `Orm`
* Getting rid of some redundant usages of the question-mark operator
* Getting rid of some unnecessary slices
* Initializing structs in a way that pleases clippy
* Removing an unnecessary `transmute` call
* Remove unnecessary `#[no_mangle]` attributes

### DIFF
--- a/fog/ocall_oram_storage/trusted/src/lib.rs
+++ b/fog/ocall_oram_storage/trusted/src/lib.rs
@@ -550,11 +550,8 @@ mod helpers {
 
 // This stuff must match edl file
 extern "C" {
-    #[no_mangle]
     fn allocate_oram_storage(count: u64, data_size: u64, meta_size: u64, id: *mut u64);
-    #[no_mangle]
     fn release_oram_storage(id: u64);
-    #[no_mangle]
     fn checkout_oram_storage(
         id: u64,
         idx: *const u64,
@@ -564,8 +561,6 @@ extern "C" {
         metabuf: *mut u64,
         metabuf_size: usize,
     );
-
-    #[no_mangle]
     fn checkin_oram_storage(
         id: u64,
         idx: *const u64,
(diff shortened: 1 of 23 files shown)


fog/ocall_oram_storage/trusted/src/lib.rs
```diff
@@ -550,11 +550,8 @@ mod helpers {
 
 // This stuff must match edl file
 extern "C" {
-    #[no_mangle]
     fn allocate_oram_storage(count: u64, data_size: u64, meta_size: u64, id: *mut u64);
```
garbageslam, line 553: Is it actually okay to remove no mangle here? I thought that is required for linkage to work. Is it implied by extern C or something?

fog/ocall_oram_storage/untrusted/src/lib.rs
```diff
@@ -170,7 +170,7 @@ pub unsafe extern "C" fn allocate_oram_storage(
         data_size as usize,
         meta_size as usize,
     ));
-    let id: u64 = core::mem::transmute(Box::into_raw(result));
+    let id = Box::into_raw(result) as u64;
```
garbageslam, line 173: Thank you!

fog/ocall_oram_storage/trusted/src/lib.rs
eranrund: Yup, I believe it is not needed: https://doc.rust-lang.org/std/keyword.extern.html

fog/ocall_oram_storage/untrusted/src/lib.rs
eranrund: Thank clippy 😛

garbageslam (approved):

mfaulk (approved):

PR #71, opened 2021-05-03
fog omits compressed commitment from TxOutRecord

cbeck88:
This updates TxOutRecord in a few ways:
* amount_compressed_commitment_data is now optional, and omitted
  except for backwards compat. (If backwards compat is not needed,
  then that field could be deleted entirely.)
* amount_compressed_commitment_data_crc32 is now present when
  that field is omitted. This is an IEEE crc32 of the omitted data.
* The sample paykit is updated so that it attempts to reconstruct
  the compressed commitment using the view private key. (This was
  koe's idea.) Then it checks the crc32 to confirm correct recomputation.
  This saves ~30 bytes in fog view, which we can use for the memo.
* A method is added to construct TxOutRecord more nicely, and then
  a bunch of test code and the enclave code is simplified.

### DIFF
--- a/fog/api/proto/view.proto
+++ b/fog/api/proto/view.proto
@@ -204,20 +204,6 @@ enum TxOutSearchResultCode {
     RateLimited = 5;
 }
 
-
-/// A Redacted Fog Transaction Output.
-/// This is the same as a normal TxOut, except that the fog hint is removed after processing, to save storage.
-message FogTxOut {
-    /// Amount.
-    external.Amount amount = 1;
-
-    /// Public key.
-    external.CompressedRistretto target_key = 2;
-
-    /// Public key.
-    external.CompressedRistretto public_key = 3;
-}
-
 /// The schema for the decrypted TxOutSearchResult ciphertext
 /// This is the information that the Ingest enclave produces for the user about their TxOut
 ///
@@ -231,6 +217,8 @@ message FogTxOut {
 /// and that's the version of the TxOut that you should use when building a transaction.
 message TxOutRecord {
     /// The (compressed ristretto) bytes of commitment associated to amount field in the TxOut that was recovered
+    ///
+    /// Note: This field is omitted in recent versions, because it can be reconstructed by the recipient instead.
     bytes tx_out_amount_commitment_data = 1;
     /// The masked value associated to amount field in the TxOut that was recovered
     fixed64 tx_out_amount_masked_value = 2;
@@ -267,4 +255,10 @@ message TxOutRecord {
     ///
     /// Represented as seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z.
     fixed64 timestamp = 7;
+    /// The crc32 of the commitment data bytes.
+    /// This is a 4-byte IEEE crc32 of the bytes of the tx_out_amount_commitment_data bytes, which is present if
+    /// the full tx_out_amount_commitment_data is omitted.
+    /// The client can recompute the tx_out_amount_commitment from the other data that we include.
+    /// They can confirm correct recomputation by checking this crc value.
+    fixed32 tx_out_amount_commitment_data_crc32 = 8;
 }
--- a/fog/api/tests/fog_types.rs
+++ b/fog/api/tests/fog_types.rs
@@ -198,35 +198,14 @@ fn tx_out_record_round_trip() {
 
     run_with_several_seeds(|mut rng| {
         let fog_txout = fog_types::view::FogTxOut::from(&TxOut::sample(&mut rng));
-
-        let commitment_bytes: &[u8; 32] = fog_txout.amount.commitment.as_ref();
-        let test_val = fog_types::view::TxOutRecord {
-            tx_out_amount_commitment_data: commitment_bytes.to_vec(),
-            tx_out_amount_masked_value: fog_txout.amount.masked_value,
-            tx_out_target_key_data: fog_txout.target_key.as_bytes().to_vec(),
-            tx_out_public_key_data: fog_txout.public_key.as_bytes().to_vec(),
-            tx_out_global_index: rng.next_u64(),
+        let meta = fog_types::view::FogTxOutMetadata {
+            global_index: rng.next_u64(),
             block_index: rng.next_u64(),
-            timestamp: 9,
+            timestamp: rng.next_u64(),
         };
+        let test_val = fog_types::view::TxOutRecord::new(fog_txout, meta);
 
         round_trip_message::<fog_types::view::TxOutRecord, fog_api::view::TxOutRecord>(&test_val);
-    });
-}
-
-/// Test that many random instances of prosty FogTxOut round trip with protobufy
-/// FogTxOut
-#[test]
-fn fog_tx_out_round_trip() {
-    {
-        let test_val: fog_types::view::FogTxOut = Default::default();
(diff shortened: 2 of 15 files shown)


fog/api/proto/view.proto
```diff
@@ -204,20 +204,6 @@ enum TxOutSearchResultCode {
...
-    /// Amount.
-    external.Amount amount = 1;
-
-    /// Public key.
-    external.CompressedRistretto target_key = 2;
-
-    /// Public key.
-    external.CompressedRistretto public_key = 3;
-}
-
```
garbageslam, line 220: this is omitted because it's not actually used in any public facing APIs, kyle and alex thought it's better to remove it if that isn't the case since it might confuse someone

fog/api/tests/fog_types.rs
```diff
@@ -198,38 +198,17 @@ fn tx_out_record_round_trip() {
...
-        round_trip_message::<fog_types::view::FogTxOut, fog_api::view::FogTxOut>(&test_val);
-    }
-
-    run_with_several_seeds(|mut rng| {
-        let test_val = fog_types::view::FogTxOut::from(&TxOut::sample(&mut rng));
-
-        round_trip_message::<fog_types::view::FogTxOut, fog_api::view::FogTxOut>(&test_val);
-    });
-}
-
```
garbageslam, line 232: this test is removed because FogTxOut type is removed from protobuf per discussion

fog/api/proto/view.proto
```diff
@@ -267,4 +255,10 @@ message TxOutRecord {
     ///
     /// Represented as seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z.
     fixed64 timestamp = 7;
+    /// The crc32 of the commitment data bytes.
+    /// This is a 4-byte IEEE crc32 of the bytes of the tx_out_amount_commitment_data bytes, which is present if
+    /// the full tx_out_amount_commitment_data is omitted.
+    /// The client can recompute the tx_out_amount_commitment from the other data that we include.
+    /// They can confirm correct recomputation by checking this crc value.
+    fixed32 tx_out_amount_commitment_data_crc32 = 8;
```
garbageslam, line 263: this change to the proto is the main point of the change -- the fog ingest enclave will only set this crc32 field and leave the earlier commtiment_data field empty.

garbageslam: the crc32 is also possibly not necessary, but kyle suggested that we should retain this in order to keep the same level of "checkability" as before, and this only adds a few bytes

fog/fog_types/src/view.rs
```diff
@@ -190,69 +193,178 @@ pub struct TxOutRecord {
...
+        // Now we can rebuild the Amount object from the value and shared secret
+        let amount = Amount::new(value, &tx_out_shared_secret)?;
+
+        // Check that the crc32 of amount compressed commitment matches
+        if self.amount_commitment_data_crc32
+            != crc32::checksum_ieee(amount.commitment.point.as_bytes())
+        {
+            return Err(FogTxOutError::ChecksumMismatch);
         }
+
```
garbageslam, line 326: this shows how reconstructing the commitment and checking the crc32 value works

fog/fog_types/src/view.rs
```diff
@@ -190,69 +193,178 @@ pub struct TxOutRecord {
...
+pub struct FogTxOutMetadata {
+    /// The global index of this TxOut in the set of all TxOut's in the
+    /// blockchain
+    pub global_index: u64,
+    /// The index of the block in which this TxOut appeared
+    pub block_index: u64,
+    /// The timestamp of the block in which this TxOut appeared, in seconds
+    /// since the Unix epoch.
+    pub timestamp: u64,
+}
```
garbageslam, line 370: this is added in order to have a nice constructor function for TxOutRecord, otherwise you take three different u64's as arguments and if they are in the wrong order, you have a bug

fog/ingest/enclave/impl/tests/tx_processing.rs
```diff
@@ -82,7 +82,7 @@ fn test_ingest_enclave(logger: Logger) {
         assert_eq!(tx_rows.len(), 10);
 
         // Check that the tx row ciphertexts have the right size
-        const EXPECTED_PAYLOAD_SIZE: usize = 188; // The observed tx_row.payload size
+        const EXPECTED_PAYLOAD_SIZE: usize = 159; // The observed tx_row.payload size
```
garbageslam, line 85: this shows the change in size of TxOutRecord when it is serialized, so we saved 29 bytes

fog/api/proto/view.proto
```diff
@@ -231,6 +217,8 @@ message FogTxOut {
 /// and that's the version of the TxOut that you should use when building a transaction.
 message TxOutRecord {
     /// The (compressed ristretto) bytes of commitment associated to amount field in the TxOut that was recovered
+    ///
+    /// Note: This field is omitted in recent versions, because it can be reconstructed by the recipient instead.
```
eranrund, line 221: You can annotate it as deprecated: see `deprecated` on https://developers.google.com/protocol-buffers/docs/proto3

fog/fog_types/src/view.rs
```diff
@@ -190,69 +193,178 @@ pub struct TxOutRecord {
...
+            tx_out_global_index: meta.global_index,
+            block_index: meta.block_index,
+            timestamp: meta.timestamp,
+        }
+    }
+
+    /// Extract a FogTxOut object from the (flattened) TxOutRecord object
+    /// Note that this discards some metadata (timestamp, block_index,
+    /// global_index).
     pub fn get_fog_tx_out(&self) -> Result<FogTxOut, KeyError> {
```
eranrund, line 228: I believe this should be `TryInto`

eranrund (approved):
LGTM

fog/api/proto/view.proto
garbageslam: for running deployments of fog, though, we can't updated these bytes. so whether the field can be deprecated depends on if those fogs will be started after beta or something like this. otherwise, sometimes people will have to keep reading this field

garbageslam:
This passed deploy

voloshyn (approved):

fog/api/proto/view.proto
garbageslam: after further discussion, we cannot deprecate this field, we believe it will continue to exist in the wild

kylefleming (approved):
Proto interface LGTM.

garbageslam:
rebased on master

fog/sample-paykit/src/cached_tx_data/mod.rs
```diff
@@ -582,7 +582,10 @@ impl OwnedTxOut {
     // so we need to backport that into the rust sample paykit.
     pub fn new(rec: TxOutRecord, account_key: &AccountKey) -> StdResult<Self, TxOutMatchingError> {
         // Reconstitute FogTxOut from the "flattened" data in TxOutRecord
-        let tx_out = rec.get_fog_tx_out()?;
+        let fog_tx_out = rec.get_fog_tx_out()?;
+
+        // Reconstute TxOut from FogTxOut and our view private key
+        let tx_out = fog_tx_out.try_recover_tx_out(&account_key.view_private_key())?;
```
garbageslam, line 588: this shows where the sample paykit is using the new code

PR #90, opened 2021-06-11
Adjust fog ledger proto

cbeck88:
The fog ledger proto contains a message called Block, but this
also exists in the blockchain.proto. This causes a naming conflict
when building go bindings for the grpc bridge.

The simplest way to resolve this issue is to rename the fog proto.
In this commit, we rename it to `BlockData` and update the fog ledger
server for this.

This is not a breaking change for clients because renaming a message
doesn't break the protobuf API. When they upgrade to the new proto
version though, they might have to update the names of some generated
structs.

### DIFF
--- a/fog/api/proto/ledger.proto
+++ b/fog/api/proto/ledger.proto
@@ -171,14 +171,14 @@ message BlockRequest {
 
 message BlockResponse {
     /// The block data returned by the server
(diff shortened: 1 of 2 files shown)


voloshyn (approved):
LGTM

eranrund (approved):

PR #95, opened 2021-06-17
Revert "Revert "fog omits compressed commitment from TxOutRecord (#71)" (#75)"

cbeck88:
This reverts commit 8b97079.

This re-applies that commit after the release branch has been created
and we have agreed to release this commit in the next release

Note that this is a breaking change for the SDKs

### DIFF
--- a/fog/api/tests/fog_types.rs
+++ b/fog/api/tests/fog_types.rs
@@ -198,35 +198,14 @@ fn tx_out_record_round_trip() {
 
     run_with_several_seeds(|mut rng| {
         let fog_txout = fog_types::view::FogTxOut::from(&TxOut::sample(&mut rng));
-
-        let commitment_bytes: &[u8; 32] = fog_txout.amount.commitment.as_ref();
-        let test_val = fog_types::view::TxOutRecord {
-            tx_out_amount_commitment_data: commitment_bytes.to_vec(),
-            tx_out_amount_masked_value: fog_txout.amount.masked_value,
-            tx_out_target_key_data: fog_txout.target_key.as_bytes().to_vec(),
-            tx_out_public_key_data: fog_txout.public_key.as_bytes().to_vec(),
-            tx_out_global_index: rng.next_u64(),
+        let meta = fog_types::view::FogTxOutMetadata {
+            global_index: rng.next_u64(),
             block_index: rng.next_u64(),
-            timestamp: 9,
+            timestamp: rng.next_u64(),
         };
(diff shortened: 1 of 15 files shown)


voloshyn (approved):
LGTM

fog/api/tests/fog_types.rs
```diff
@@ -198,38 +198,17 @@ fn tx_out_record_round_trip() {
...
         let fog_txout = fog_types::view::FogTxOut::from(&TxOut::sample(&mut rng));
-
-        let commitment_bytes: &[u8; 32] = fog_txout.amount.commitment.as_ref();
-        let test_val = fog_types::view::TxOutRecord {
-            tx_out_amount_commitment_data: commitment_bytes.to_vec(),
-            tx_out_amount_masked_value: fog_txout.amount.masked_value,
-            tx_out_target_key_data: fog_txout.target_key.as_bytes().to_vec(),
-            tx_out_public_key_data: fog_txout.public_key.as_bytes().to_vec(),
-            tx_out_global_index: rng.next_u64(),
+        let meta = fog_types::view::FogTxOutMetadata {
```
tubi70, line 201: Suggest adding a comment here for what FogTxOutMetaData

fog/fog_types/src/view.rs
```diff
@@ -190,69 +193,178 @@ pub struct TxOutRecord {
...
+    ///   and isn't needed by the client.
+    ///
+    /// Arguments:
+    /// * view_key: the private view key of the recipient of this TxOut
+    ///
+    /// Returns:
+    /// * TxOut,
+    /// Or
+    /// * An error if recovery failed
+    pub fn try_recover_tx_out(&self, view_key: &RistrettoPrivate) -> Result<TxOut, FogTxOutError> {
```
tubi70, line 306: Suggest renaming method to "try_recover_tx_out_fogtxtout" which would provide a more descriptive name to understand the functionality

fog/api/tests/fog_types.rs
garbageslam: there was a comment like this in the (original) PR: https://github.com/mobilecoinfoundation/fog/pull/71#discussion_r625366668

we could add a code comment, it would say something like "Gather metadata for FogTxOut" ? does that sound good to you?

fog/fog_types/src/view.rs
garbageslam: that name seems repetitive to me -- this is a member function on `FogTxOut` object already, so the function can already be invoked like `FogTxOut::try_recover_tx_out`

tubi70 (approved):

PR #98, opened 2021-06-25
fix fog view attest error status code

cbeck88:
this is supposed to be "permission denied" so that the clients
can know that they should retry by trying to make a new attested
connection. fog-ledger was already doing this correctly but
it turns out that in this case fog-view wasn't


sugargoat (approved):

jcape (approved):
LGTM

PR #109, opened 2021-07-15
fog-report-cli improvements

cbeck88:
This makes a second "mode" for fog-report-cli where it doesn't
do any validation of the fog report response and just parses the
data of interest out of it -- that is, the hex bytes of the fog
ingress pubkey, and its pubkey expiry value.

This is useful because sometimes, when the system gets stuck,
the users report "tombstone block exceeded" and it can be traced
back to the pubkey expiry value in the fog report server.

It's also useful because it lets us quickly see the most recent
fog ingress pubkey that has been published. If the active server
dies, knowing the hex bytes of this key may help you figure out
which backup to restart, if they have different keys.

Some pitfalls with this workflow, before this commit:

(1) You had to make a pubkey file which has all the right data,
    which requires going to another repo, building, and running
    a tool to generate keys, and then getting the artifacts to
    this repo and pointing this program at the keyfiles, which
    is cumbersome.
    In the new revision, you can specify fog url and SPKI as
    command-line args and skip the keyfile stuff.
(2) If you don't download all the correct measurements and SPKI's
    for the network, and use that SPKI when you generate sample
    keys, the attestation check or the fog signature check will
    fail and it takes some time to work through this.
    If you are just trying to diagnose why a network is stopped,
    this is kind of a time suck.
    In the new revision, if there is no public address file and
    no spki provided, (only fog url), then we skip the normal
    validation pathway, and use the mc-attest-core types to
    parse the verification report (without validating it) and get
    the data of interest, without any possibility of failure due
    to IAS, attestation, certificate chains, and so on.

Ideally, there would be some kind of dashboard that can provide
this and other information without relying on a cli, but this is
a decent starting point, and may help us diagnose problems later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1617,6 +1617,7 @@ dependencies = [
 name = "fog-report-cli"
 version = "1.1.0"
 dependencies = [
+ "base64 0.12.3",
  "binascii",
  "fog-ingest-enclave-measurement",
  "grpcio",
@@ -1626,6 +1627,7 @@ dependencies = [
  "mc-crypto-keys",
  "mc-fog-report-connection",
  "mc-fog-report-validation",
+ "mc-sgx-types",
  "mc-util-keyfile",
  "mc-util-uri",
  "structopt",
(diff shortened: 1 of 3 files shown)


joekottke (approved):

sugargoat (approved):

eranrund:
I think we should validate by default and there should be a `--skip-validate` to explicitly disable that behavior. @jcape might have thoughts on this as well.

garbageslam:
okay, i will open a second PR. but this is mostly only useful in tests and diagnostics.

the only time it doesn't validate is if there is no authority sig or spki, and so validation would be impossible

eranrund:
My thinking is that people will not pass the authority sig or spki since they are optional and not go through the thought process of "oh, i'm not providing that, no validation is taking place".

garbageslam:
yeah, okay

PR #110, opened 2021-07-15
Add --no-validate flag to fog-report-cli

cbeck88:
After discussion we think this might be easier to use correctly


eranrund (approved):

eranrund:
Thank you!

sugargoat (approved):

PR #116, opened 2021-07-29
fix FOG-180: enclave report cache updated too frequently

cbeck88:
this commit attempts to make it so that ingest server updates
the enclave report cache only when necessary

This is:
(1) At least once before anything happens
(2) Whenever the ingress key has changed
(3) Whenever we notice that the report doesn't match the key
    we think we are publishing to the database.

Additionally, a background thread is created that updates it
at least as frequently as consensus (18 hours)

Before this commit, we update it at least once per block.

There are some questions about whether DNS failures with looking
for IAS were causing instability in the service.
This commit doesn't speak directly to that, but it
can't hurt to hit up IAS less often than we were doing.


fog/ingest/server/src/controller.rs
```diff
@@ -1028,11 +1071,34 @@ where
         ingress_public_key: &CompressedRistrettoPublic,
         state: &mut MutexGuard<IngestControllerState>,
     ) -> Result<IngressPublicKeyStatus, Error> {
-        self.update_enclave_report_cache()?;
```
garbageslam, line 1031: removing this line, or rather, making it conditional on the current report not matching the key we are supposed to be publishing, is the main point of this PR

fog/ingest/server/src/worker.rs
```diff
@@ -275,3 +288,73 @@ impl Drop for PeerCheckupWorker {
         }
     }
 }
+
+/// The report cache worker is a thread responsible for periodically calling
+/// update report cache. This is a separate thread so that it can be on a
+/// time-based schedule, so it will happen even if there are few blocks.
+pub struct ReportCacheWorker {
```
garbageslam, line 295: this worker object is added for fog-ingest, and it is not exactly the same as the one in `mc_sgx_report_cache_untrusted`. 

It can't be because, while in consensus, the "enclave identity" bytes never change, in fog ingest, they do, because sometimes the ingress key changes, and then the bytes in the report have to change.

(1) Sometimes, the operator calls "new_keys"
(2) Sometimes, the active node tells us "hey this is the key"
(3) Sometimes, we reach out to a remote node and say "hey give me your key" (if the operator asks us too)

Before this PR, we would get an updated report, whenever we were about to take an action involving reports or attestation.

After this PR, we update the report
* once when the server starts
* every 18 hours according to background thread, same as consensus
* after actions 1, 2, 3 listed above
* before some miscellaneous actions: activating the server, syncing key from remote. these two things happen very rarely (with a human in the loop) so it should be fine, and keeping these prevents us having to refactor tests.

fog/ingest/server/src/worker.rs
```diff
@@ -254,14 +255,26 @@ impl PeerCheckupWorker {
...
+                let mut last_refreshed_at = Instant::now();
+
+                loop {
+                    if stop_requested.load(Ordering::SeqCst) {
+                        log::debug!(logger, "Peer checkup thread stop requested.");
+                        break;
+                    }
 
-                controller.peer_checkup();
-                std::thread::sleep(peer_checkup_period);
```
garbageslam, line 264: this change was tangential to the goal of the PR, but i realized that this thread is sleeping for like 60 seconds. that probably makes the tests take at least 60 seconds. so if we make this only sleep 1 second at a time, same as consensus thread, this should make the unit tests run a lot faster.

fog/ingest/server/src/controller.rs
```diff
@@ -1336,3 +1403,19 @@ where
         Ok(())
     }
 }
+
+// Helper function which extracts ingress pubkey bytes from (our own) report, to
+// double check
+fn try_extract_ingress_pubkey_from_report(
```
garbageslam, line 1409: this code is mostly the same as what we added to fog-report-cli last week-ish

fog/ingest/server/src/controller.rs
```diff
@@ -782,8 +830,16 @@ where
         log::info!(self.logger, "Key successfully set in enclave: {}", pubkey);
 
         *self.last_sealed_key.lock().unwrap() = None;
-        self.write_state_file();
-        Ok(self.get_ingest_summary())
+        self.write_state_file_inner(&mut state);
+        let result = self.get_ingest_summary_inner(&mut state);
+
+        // Don't hold the state mutex while we are talking to IAS
```
garbageslam, line 836: i don't know that this is really critical, but it seems like a good idea and can't hurt

fog/ingest/server/src/controller.rs
```diff
@@ -1028,11 +1081,34 @@ where
...
         state: &mut MutexGuard<IngestControllerState>,
     ) -> Result<IngressPublicKeyStatus, Error> {
-        self.update_enclave_report_cache()?;
+        // Get a report and check that it makes sense with what we think is happening
+        let report = {
+            let report = self.enclave.get_ias_report()?;
+            // Check that key in report data matches ingress_public_key.
+            // If not, then there is some kind of race.
+            let found_key = try_extract_ingress_pubkey_from_report(&report)?;
+            if &found_key == ingress_public_key {
```
garbageslam, line 1090: this is some defensive programming:
even if our code doesn't actually have a race here today, that doesn't mean that it won't in the future.

if this condition fails, and we publish a report to the fog db that doesn't actually match the public key, that is likely catastrophic because the users will encrypt for a key which isn't the same as what the enclave is decrypting for, and it will be very complicated to mitigate that after the fact, and possibly require the users to download and scan a very large amount of data. so just double checking and refusing to publish if we can't fix the mismatch seems like the best thing to do.

fog/ingest/server/src/controller.rs
```diff
@@ -222,8 +259,7 @@ where
     pub fn new_keys(&self) -> Result<IngestSummary, Error> {
         let mut state = self.get_state();
         self.new_keys_inner(&mut state)?;
-        self.write_state_file_inner(&mut state);
-        Ok(self.get_ingest_summary())
+        Ok(self.get_ingest_summary_inner(&mut state))
```
garbageslam, line 262: this was likely a bug -- this function was calling `get_ingest_summary` while holding the state lock mutex, so i would expect a deadlock

fog/ingest/server/src/worker.rs
sugargoat: Thank you for the explanation here!

fog/ingest/server/src/worker.rs
sugargoat: Good find

fog/ingest/server/src/controller.rs
```diff
@@ -1028,11 +1081,34 @@ where
...
+
+                let report = self.enclave.get_ias_report()?;
+                let found_key = try_extract_ingress_pubkey_from_report(&report)?;
+                if &found_key == ingress_public_key {
+                    report
+                } else {
+                    // This means that the caller is wrong about what the
+                    // current ingress public key is, and we don't have anything we can publish.
+                    log::error!(self.logger, "Report doesn't contain the expected public key even after report refresh: {:?} != {:?}", found_key, ingress_public_key);
+                    return Err(Error::PublishReport);
```
sugargoat, line 1104: Making a note that we will want to have enough information to take action in the error message - what is the node operator expected to do to resolve this scenario?

sugargoat (approved):
LGTM - would request a review from either @eranrund or @jcape, and for @joekottke or @jgreat to approve for understanding how they should handle the error scenario

fog/ingest/server/src/controller.rs
garbageslam: i have no idea how this would happen, we would have to investigate to know what is wrong. it would probably mean that like, some cache involved in creating the reports is in a bad state that is hard to reproduce.

if it doesn't resolve itself then maybe we could try killing the server and making a backup node active instead.

fog/ingest/server/src/controller.rs
garbageslam: there needs to be a test that exercises this API, clearly there wasn't one before this

fog/ingest/server/src/controller.rs
```diff
@@ -222,8 +259,7 @@ where
     pub fn new_keys(&self) -> Result<IngestSummary, Error> {
         let mut state = self.get_state();
         self.new_keys_inner(&mut state)?;
-        self.write_state_file_inner(&mut state);
```
garbageslam, line 225: we dont' need to write state file here, because `new_keys_inner` is doing that

fog/ingest/server/src/worker.rs
garbageslam: i guess there is one more case:
- if we ask for a report to publish, and it doesn't match the key we think we have, then we update the report and try again.

fog/ingest/server/src/worker.rs
```diff
@@ -275,3 +288,73 @@ impl Drop for PeerCheckupWorker {
...
+/// The report cache worker is a thread responsible for periodically calling
+/// update report cache. This is a separate thread so that it can be on a
+/// time-based schedule, so it will happen even if there are few blocks.
+pub struct ReportCacheWorker {
+    stop_requested: Arc<AtomicBool>,
+    thread: Option<JoinHandle<()>>,
+}
+
+impl ReportCacheWorker {
+    /// Create a new PeerCheckupWorker thread
```
garbageslam, line 301: this comment needs updating

fog/ingest/server/src/controller.rs
garbageslam: we should probably share this code in those two places, but i'm actually not sure what would be a good place to put it, and i think the fog-report-cli change doesn't exist in the release branch

garbageslam:
this passed deployment

fog/ingest/server/src/controller.rs
```diff
@@ -200,15 +205,47 @@ where
...
+        if old_ingress_pubkey != new_ingress_pubkey {
+            // Seal the new private keys we ended up with to disk
+            *self.last_sealed_key.lock().unwrap() = None;
+            self.write_state_file_inner(&mut state);
+
+            // Don't hold the lock while we make network calls to IAS
+            drop(state);
+
+            // Refresh report cache
+            if let Err(err) = self.update_enclave_report_cache() {
```
eranrund, line 241: A debug log that we are attempting to update would be nice.

fog/ingest/server/src/controller.rs
```diff
@@ -242,6 +278,15 @@ where
         *self.last_sealed_key.lock().unwrap() = None;
         self.write_state_file_inner(state);
 
+        // Update report cache
+        if let Err(err) = self.update_enclave_report_cache() {
```
eranrund, line 282: A debug log could be nice

fog/ingest/server/src/controller.rs
eranrund: might be nice to just include this as a comment

fog/ingest/server/src/controller.rs
eranrund: I agree this should not be duplicated. Maybe in the [API](https://github.com/mobilecoinfoundation/fog/tree/master/fog/api/src) crate?

fog/ingest/server/src/error.rs
```diff
@@ -53,6 +55,8 @@ pub enum IngestServiceError {
     Io(std::io::Error),
     /// GRPC Error: {0}
     Grpc(GrpcError),
+    /// Error parsing Attestation Verification Report: {0}
```
eranrund, line 58: I would shorten this to `Report parse: {0}`

fog/ingest/server/src/error.rs
```diff
@@ -127,6 +131,41 @@ impl From<GrpcError> for IngestServiceError {
...
+impl From<ReportParseError> for IngestServiceError {
+    fn from(src: ReportParseError) -> Self {
+        Self::ReportParse(src)
+    }
+}
+
+/// An error which occurs when parsing an AVR
+#[derive(Debug, Display)]
+pub enum ReportParseError {
+    /// Verification error: {0}
```
eranrund, line 143: I would drop the `error` suffix, it should be obvious when it is displayed that it is an error. Same for the other variant below.

fog/ingest/server/src/worker.rs
eranrund: Maybe instead of duplicating it we should change `mc_sgx_report_cache_untrusted` to accept a callback for getting the report data? Not something I feel strongly about, since in this case the implementation is very small.

eranrund (approved):
LGTM, although I am wondering if we should not move all report refreshing to the worker thread, and add an option to signal it to force a refresh now and not wait for the internal period to exceed. What would be the implications of moving the report refreshing to happen asynchronously that way?
The advantage of doing that would be that all IAS interaction (network calls) happens from a single thread - so it doesn't block anything, and also we are not doing two refreshes concurrently.

garbageslam:
> LGTM, although I am wondering if we should not move all report refreshing to the worker thread, and add an option to signal it to force a refresh now and not wait for the internal period to exceed. What would be the implications of moving the report refreshing to happen asynchronously that way?
The advantage of doing that would be that all IAS interaction (network calls) happens from a single thread - so it doesn't block anything, and also we are not doing two refreshes concurrently.

So james and i talked about this -- the thing is, sometimes it is important for the main thread to be able to, trigger a refresh, and then, block, until that refresh is completed, because it makes the semantics of all of this easier.

All these lines where we have "reach out to a peer and make an attested connection", fail with "no report" if the report cache hasn't been populated, and then the server crashes. Because this line: https://github.com/mobilecoinfoundation/fog/pull/116/files#diff-0f4f3e64562ef1cbd4d4d712b8f5fcb08ef44066474f57a7005059a348293064R191
blocks the main thread until the report cache succeeded, i know that none of the workers will start doing things until the cache is populated at least once.

Another thing is, it makes it easier to reason about code like this: https://github.com/mobilecoinfoundation/fog/pull/116/files#diff-1bd75fa7d4e53379b9fe76ed5de3edd920c698f1d188a08ae34d2ba9fb414247R1103

Right now, I know that if we hit this log message, it means that
(1) the server is in the active state
(2) the enclave told us earlier that the key was one thing, then it gave us a report that didn't match that
(3) even calling update_enclave_report_cache didn't fix that

so I can reason that either, (1) the enclave key is somehow being allowed to change while the server is in active state (big bug), or (2) the report cache mechanism is messed up somehow, and the caches either in the enclave or out of the enclave are not being flushed when we trigger the update (big bug).

if that line merely triggers an asynchronous update that doesn't block the worker thread, then I can't reason in that way, because it may just be that the update didn't finish yet.

> The advantage of doing that would be that all IAS interaction (network calls) happens from a single thread - so it doesn't block anything, and also we are not doing two refreshes concurrently.

it's worth drilling in on this -- in the current implementation, the `controller` holds `Arc<Mutex<ReportCache>>`: https://github.com/mobilecoinfoundation/fog/blob/5a776bd027e1e2cc5f4717f4d4f8f56f30ae0585/fog/ingest/server/src/controller.rs#L70

So this serializes all calls to `update_enclave_report_cache` no matter which thread is triggering it, and this mutex is only locked when that function is called, and that function doesn't lock any other mutexes, so there's no risk of deadlock.

fog/ingest/server/src/worker.rs
garbageslam: yeah -- i think another way is, we could make an additional constructor of that thing that takes `Arc<Mutex<ReportCache>>` so that other things can also trigger the update?

it will be easier to develop changes like this after fog is merged into mobilecoin

eranrund:
> > LGTM, although I am wondering if we should not move all report refreshing to the worker thread, and add an option to signal it to force a refresh now and not wait for the internal period to exceed. What would be the implications of moving the report refreshing to happen asynchronously that way?
> > The advantage of doing that would be that all IAS interaction (network calls) happens from a single thread - so it doesn't block anything, and also we are not doing two refreshes concurrently.
> 
> So james and i talked about this -- the thing is, sometimes it is important for the main thread to be able to, trigger a refresh, and then, block, until that refresh is completed, because it makes the semantics of all of this easier.
> 
> All these lines where we have "reach out to a peer and make an attested connection", fail with "no report" if the report cache hasn't been populated, and then the server crashes. Because this line: https://github.com/mobilecoinfoundation/fog/pull/116/files#diff-0f4f3e64562ef1cbd4d4d712b8f5fcb08ef44066474f57a7005059a348293064R191
> blocks the main thread until the report cache succeeded, i know that none of the workers will start doing things until the cache is populated at least once.
> 
> Another thing is, it makes it easier to reason about code like this: https://github.com/mobilecoinfoundation/fog/pull/116/files#diff-1bd75fa7d4e53379b9fe76ed5de3edd920c698f1d188a08ae34d2ba9fb414247R1103
> 
> Right now, I know that if we hit this log message, it means that
> (1) the server is in the active state
> (2) the enclave told us earlier that the key was one thing, then it gave us a report that didn't match that
> (3) even calling update_enclave_report_cache didn't fix that
> 
> so I can reason that either, (1) the enclave key is somehow being allowed to change while the server is in active state (big bug), or (2) the report cache mechanism is messed up somehow, and the caches either in the enclave or out of the enclave are not being flushed when we trigger the update (big bug).
> 
> if that line merely triggers an asynchronous update that doesn't block the worker thread, then I can't reason in that way, because it may just be that the update didn't finish yet.
> 
> > The advantage of doing that would be that all IAS interaction (network calls) happens from a single thread - so it doesn't block anything, and also we are not doing two refreshes concurrently.
> 
> it's worth drilling in on this -- in the current implementation, the `controller` holds `Arc<Mutex<ReportCache>>`:
> 
> https://github.com/mobilecoinfoundation/fog/blob/5a776bd027e1e2cc5f4717f4d4f8f56f30ae0585/fog/ingest/server/src/controller.rs#L70
> 
> So this serializes all calls to `update_enclave_report_cache` no matter which thread is triggering it, and this mutex is only locked when that function is called, and that function doesn't lock any other mutexes, so there's no risk of deadlock.

I see. Thank you for the detailed explanation. The ability to block does make sense in the cases you've mentioned, and since the `report_cache` mutex serializes the updates then I am satisfied.

fog/ingest/server/src/controller.rs
garbageslam: sounds good, i will put it there

garbageslam: i ran `cargo tree` in `fog-api` crate, it already depends on `mc-attest-ake` and `mc-attest-core` anyways

fog/api/src/extra.rs
```diff
@@ -0,0 +1,60 @@
+use core::convert::TryFrom;
```
queenbdubs, line 1: why is this file called `extra.rs`?

garbageslam: this is this like, one off function that is getting the key out of the report and isn't validating anything, so in that sense like, clients aren't really supposed to use it.

I don't know, can you think of a better name?

fog/ingest/server/src/worker.rs
queenbdubs: related question...why did you make a new ReportCacheWorker instead of just using the existing ReportCacheThread?  I see that you based the ReportCacheWorker on a PeerCheckupWorker--what does a PeerCheckupWorker do?

fog/api/src/extra.rs
queenbdubs: Oh I see.  I thought maybe you had mistyped and it was supposed to be `extract.rs` so what about that?

queenbdubs (approved):
At one point, we (I + someone...either you/sara/eran?) considered also keeping track of the last block where the report cache was updated.  Would this still be useful to do?  I imagine it would be useful for debugging, but I'm not sure.

fog/api/src/extra.rs
garbageslam: so actually, how about `report_parse.rs`? we called the error `ReportParseError` so at least that would match

fog/ingest/server/src/worker.rs
garbageslam: so what happened here actually was, I looked at what ReportCacheThread does, and I copied it here, but made it just call `controller.update_enclave_report_cache`, because I still want the Controller to be able to trigger update sometimes, and so I want the controller to own the cache.

The philosophy I'm trying to use here is, there's one object that establishes invariants, owns most of the state, and has entrypoints where functionality can be called, that is like, not threaded and potentially unit-testable. And that's the controller. And all the threads, grpc routes, worker threads, etc. are outside that object, and call into it when they need something done. So that thing does have to have mutexes, but it can be done in a thread-safe way, and it means that to figure out the high level behavior of the server, if there is anyway it can deadlock etc., we mostly just have to read `controller.rs`. There's some logic in the `worker.rs` related to timings and such that's relevant, but anything where they are like changing the state of the system or talking to the enclave is going through the controller at some point.

fog/ingest/server/src/controller.rs
garbageslam: done

garbageslam:
@queenbdubs I think we should make a ticket to follow up on that. It sounds like a good idea but I'm not sure if we should try to do it at this time in the release branch

eranrund:
Is anything blocking this from getting merged?

garbageslam:
@eranrund Sara wanted that ops signs off on, that they understand the change, and that they understand what should happen if they see this error about report not matching the key in a loop or something. We are doing a meeting about this at this moment

garbageslam:
TODO:
- [ ] add a metric for the pubkey_expiry of the last report we published

jcape (commented):
Some minor nitpicks that will simplify the dependencies of this (i.e. no need for sgx-types)---could you also add an [issue in mobilecoin](https://github.com/mobilecoinfoundation/mobilecoin/issues/new) for what needs to happen in ReportCache?

I'd also recommend a fog ticket for refactoring the enclave method to return bool.

fog/api/src/report_parse.rs
```diff
@@ -0,0 +1,62 @@
...
+/// FullyValidatedFogPubkey.
+pub fn try_extract_unvalidated_ingress_pubkey_from_fog_report(
+    report: &VerificationReport,
+) -> Result<CompressedRistrettoPublic, ReportParseError> {
+    let verification_report_data = VerificationReportData::try_from(report)?;
+    // This extracts the user-data attached to the report, which is a thin wrapper
+    // around [u8; 64]
+    let report_data = verification_report_data.quote.report_body()?.report_data();
+    // Unwrap the wrapper
+    let report_data = sgx_report_data_t::from(report_data);
```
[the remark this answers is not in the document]
garbageslam: thanks, didn't realize this

jcape (approved):
LGTM!

PR #119, opened 2021-08-08
Migrate from deprecated aes-ctr crate to aes crate with ctr feature

garbageslam:


jcape (approved):
LGTM, but Cargo.lock needs merging

eranrund (approved):

PR #122, opened 2021-08-11
update fog and fog clients for encrypted memo changes: recoverable transaction history

cbeck88:
This makes fog ingest include the encrypted memo in the TxOutRecord
object. It then passes through fog view as usual. We also update
the FogTxOut object to parse the memo out of the view response.

fog sample paykit is updated to both produce memos on new transactions,
and parse memos from incoming transactions, and remember the last one or an error.

This uprevs mobilecoin submodule, and makes the android bindings
and libmobilecoin able to build.

It installs a RTHMemoBuilder when those bindings create transaction
builder, but it does not configure the RTHMemoBuilder, because
we probably need to extend the bindings to do that properly.

It also gets all the tests passing

Still TODO:
- [x] Make sample paykit actually read the memos and expose this info in some basic way
- [x] Make fog test client confirm that the memos are working (if A sends to B then B got the right memo and so did A)

### DIFF
--- a/fog/ingest/enclave/impl/tests/tx_processing.rs
+++ b/fog/ingest/enclave/impl/tests/tx_processing.rs
@@ -82,7 +82,7 @@ fn test_ingest_enclave(logger: Logger) {
         assert_eq!(tx_rows.len(), 10);
 
         // Check that the tx row ciphertexts have the right size
-        const EXPECTED_PAYLOAD_SIZE: usize = 159; // The observed tx_row.payload size
+        const EXPECTED_PAYLOAD_SIZE: usize = 207; // The observed tx_row.payload size
         for tx_row in tx_rows.iter() {
             assert_eq!(
                 tx_row.payload.len(), EXPECTED_PAYLOAD_SIZE,
--- a/fog/sample-paykit/src/cached_tx_data/memo_handler.rs
+++ b/fog/sample-paykit/src/cached_tx_data/memo_handler.rs
@@ -4,6 +4,7 @@
 
 use displaydoc::Display;
 use mc_account_keys::{AccountKey, AddressHash, PublicAddress, CHANGE_SUBADDRESS_INDEX};
+use mc_common::logger::{log, Logger};
 use mc_crypto_keys::{KeyError, RistrettoPublic};
 use mc_transaction_core::{get_tx_out_shared_secret, subaddress_matches_tx_out, tx::TxOut};
 use mc_transaction_std::{MemoDecodingError, MemoType};
@@ -19,17 +20,19 @@ use std::{collections::HashMap, convert::TryFrom};
 pub struct MemoHandler {
     contacts: HashMap<AddressHash, PublicAddress>,
     last_memo: Result<Option<MemoType>, MemoHandlerError>,
+    logger: Logger,
 }
 
 impl MemoHandler {
     /// Make a new memo handler with a given set of contacts
-    pub fn new(address_book: Vec<PublicAddress>) -> Self {
+    pub fn new(address_book: Vec<PublicAddress>, logger: Logger) -> Self {
         Self {
             contacts: address_book
                 .into_iter()
                 .map(|addr| (AddressHash::from(&addr), addr))
                 .collect(),
             last_memo: Ok(None),
+            logger,
         }
     }
 
@@ -57,6 +60,7 @@ impl MemoHandler {
         let memo_payload = tx_out.decrypt_memo(&shared_secret);
 
         let memo_type = MemoType::try_from(&memo_payload)?;
+        log::info!(self.logger, "Obtained a memo: {:?}", memo_type);
         match memo_type.clone() {
             MemoType::Unused(_) => Ok(None),
             MemoType::AuthenticatedSender(memo) => {
--- a/fog/sample-paykit/src/cached_tx_data/memo_handler.rs
+++ b/fog/sample-paykit/src/cached_tx_data/memo_handler.rs
@@ -1,0 +1,132 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+//! A memo handler object which processes memos, for use in integration tests
+
+use displaydoc::Display;
+use mc_account_keys::{AccountKey, AddressHash, PublicAddress, CHANGE_SUBADDRESS_INDEX};
+use mc_crypto_keys::{KeyError, RistrettoPublic};
+use mc_transaction_core::{get_tx_out_shared_secret, subaddress_matches_tx_out, tx::TxOut};
+use mc_transaction_std::{MemoDecodingError, MemoType};
+use std::{collections::HashMap, convert::TryFrom};
+
+/// A handler object that holds a contacts list and tries to recieve and
+/// authenticate memos. It provides the "get_last_memo" function which can be
+/// used to check what the last memo recieved was, and if there were validation
+/// errors.
+///
+/// This is useful for test code.
+#[derive(Debug, Clone)]
+pub struct MemoHandler {
+    contacts: HashMap<AddressHash, PublicAddress>,
+    last_memo: Result<Option<MemoType>, MemoHandlerError>,
+}
+
+impl MemoHandler {
+    /// Make a new memo handler with a given set of contacts
+    pub fn new(address_book: Vec<PublicAddress>) -> Self {
+        Self {
+            contacts: address_book
+                .into_iter()
+                .map(|addr| (AddressHash::from(&addr), addr))
+                .collect(),
+            last_memo: Ok(None),
+        }
+    }
+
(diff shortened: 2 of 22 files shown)


libmobilecoin/src/common/boundary.rs
```diff
@@ -52,11 +52,25 @@ where
...
+    //
+    // The reason we use Box<dyn + ...> at all is to avoid making everything
+    // a generic parameter of transaction builder, which multiplies the number of
+    // types that might have to have cross-language bindings.
+    //
+    // UnwindSafe is too restrictive -- the goal of UnwindSafe is that if a panic is
+    // caught, we cannot "easily" observe a broken invariant. However, the only
+    // thing we actually need at an ffi boundary is to prevent unwinding across
+    // stackframes into swift etc.
+    catch_unwind(AssertUnwindSafe(f))
```
garbageslam, line 73: @voloshyn @mobilecoinfoundation/sdk Adam Mork, 

let me know if you guys agree with the changes in this file, I think this is the right thing to do (remove the `UnwindSafe` bound that was on all of these functions), I have tried to explain why I did this.

If I don't do this, I get build errors like this:

```
error[E0277]: the type `(dyn MemoBuilder + Send + Sync + 'static)` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
   --> libmobilecoin/src/transaction.rs:343:5
    |
343 |     ffi_boundary_with_error(out_error, || {
    |     ^^^^^^^^^^^^^^^^^^^^^^^ `(dyn MemoBuilder + Send + Sync + 'static)` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
    | 
   ::: libmobilecoin/src/common/boundary.rs:38:15
    |
38  | pub(crate) fn ffi_boundary_with_error<R, I>(
    |               ----------------------- required by a bound in this
39  |     out_error: FfiOptMutPtr<FfiOptOwnedPtr<McError>>,
40  |     f: impl (FnOnce() -> Result<R, LibMcError>) + UnwindSafe,
    |                                                   ---------- required by this bound in `boundary::ffi_boundary_with_error`
    |
    = help: within `std::option::Option<TransactionBuilder<FogResolver>>`, the trait `RefUnwindSafe` is not implemented for `(dyn MemoBuilder + Send + Sync + 'static)`
    = note: required because it appears within the type `*const (dyn MemoBuilder + Send + Sync + 'static)`
    = note: required because it appears within the type `Unique<(dyn MemoBuilder + Send + Sync + 'static)>`
    = note: required because it appears within the type `Box<(dyn MemoBuilder + Send + Sync + 'static)>`
    = note: required because it appears within the type `std::option::Option<Box<(dyn MemoBuilder + Send + Sync + 'static)>>`
    = note: required because it appears within the type `TransactionBuilder<FogResolver>`
    = note: required because it appears within the type `std::option::Option<TransactionBuilder<FogResolver>>`
    = note: required because of the requirements on the impl of `UnwindSafe` for `mc_util_ffi::FfiMutPtr<'_, std::option::Option<TransactionBuilder<FogResolver>>>`
    = note: required because it appears within the type `[closure@libmobilecoin/src/transaction.rs:343:40: 383:6]`

```

which I think cannot be resolved if we want to have `Box<dyn ...>` in the transaction builder, which I think we do.

garbageslam:
this has passed CD

eranrund:
> this has passed CD

Hows grafana looking?

eranrund:
> We don't update the fog sample paykit to actually read the memos,
> but it does use the RTH memo build.

Is that not fog/sample-paykit/src/cached_tx_data/memo_handler.rs?

fog/sample-paykit/src/cached_tx_data/memo_handler.rs
```diff
@@ -0,0 +1,136 @@
...
+#[derive(Debug, Clone)]
+pub struct MemoHandler {
+    contacts: HashMap<AddressHash, PublicAddress>,
+    last_memo: Result<Option<MemoType>, MemoHandlerError>,
+    logger: Logger,
+}
+
+impl MemoHandler {
+    /// Make a new memo handler with a given set of contacts
+    pub fn new(address_book: Vec<PublicAddress>, logger: Logger) -> Self {
```
eranrund, line 28: Nit: `&[PublicAddress]`?

fog/sample-paykit/src/cached_tx_data/memo_handler.rs
```diff
@@ -0,0 +1,136 @@
...
+    FailedHmacValidation,
+
+    /// Failed subaddress validation
+    FailedSubaddressValidation,
+
+    /// Key: {0}
+    Key(KeyError),
+
+    /// Memo: {0}
+    Memo(MemoDecodingError),
```
eranrund, line 123: MemoDecode?

fog/test-client/src/test_client.rs
```diff
@@ -312,6 +322,57 @@ impl TestClient {
                 tgt_balance + self.transfer_amount,
             )?;
 
+            // Ensure source client got a destination memo, as expected for recoverable
```
eranrund, line 325: Nice

eranrund (approved):
Looks good as far as I can tell, although I am wondering about the decision to change the SDK bindings to use a partially-configured RTHMemoBuilder instead of the default one.

garbageslam:
I need to update the PR comment, the first commit didn't have the memo handler

garbageslam:
> Looks good as far as I can tell, although I am wondering about the decision to change the SDK bindings to use a partially-configured RTHMemoBuilder instead of the default one.

the thing is that to fully configure it, you need a `SenderMemoCredential` so that it can produce the hmac value, but to do that, we need the account key in scope probably, so we would have to change the bindings to do that. so i figured it's better to just hold off on that for now

garbageslam:
@eranrund i will post grafana graphs shortly

fog/ingest/enclave/impl/tests/tx_processing.rs
```diff
@@ -82,7 +82,7 @@ fn test_ingest_enclave(logger: Logger) {
         assert_eq!(tx_rows.len(), 10);
 
         // Check that the tx row ciphertexts have the right size
-        const EXPECTED_PAYLOAD_SIZE: usize = 159; // The observed tx_row.payload size
+        const EXPECTED_PAYLOAD_SIZE: usize = 207; // The observed tx_row.payload size
```
garbageslam, line 85: @jcape this is the new size of e txout record

fog/sample-paykit/src/cached_tx_data/memo_handler.rs
garbageslam: i guess i prefer to take it by value, since it avoids a copy? why do you have this nit?

libmobilecoin/src/common/boundary.rs
garbageslam: @the-real-adammork ^

fog/sample-paykit/src/cached_tx_data/memo_handler.rs
eranrund: Disregard, I missed the part where it grabs `addr`.

jcape (approved):
LGTM FWIW.

I'm a bit concerned about catch unwind changes---like, it seems like the `jni` crate gets around this kind of thing by just going ahead and wrapping everything in `Arc<Mutex<Foo>>` and passing pointers to that, and is just highlighting that we're letting the freak flag fly (in a bad way) here.

It's not for this PR, of course, but maybe worth looking into as safety improvement work.

garbageslam:
@jcape the error says this:

```
error[E0277]: the type `(dyn MemoBuilder + Send + Sync + 'static)` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
   --> libmobilecoin/src/transaction.rs:343:5
    |
343 |     ffi_boundary_with_error(out_error, || {
    |     ^^^^^^^^^^^^^^^^^^^^^^^ `(dyn MemoBuilder + Send + Sync + 'static)` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
    | 
```

So is `Arc<Mutex<X>>` going to fix that? `Arc<Mutex<X>>` can also contain interior mutability I think

garbageslam:
@jcape i have read some more about this -- i think the best alternative I see here is, add an `UnwindSafe` trait bound to the `MemoBuilder` constraints in `TransactionBuilder`, to allow that people can use `catch_unwind` around transaction builder.

I don't think we should use `Arc<Mutex<MemoBuilder>>`, it just seems like an overkill

garbageslam:
https://github.com/rust-lang/rust/issues/40628#issue-215168677

samdealy (approved):

the-real-adammork (approved):

PR #123, opened 2021-08-12
initial change to consolidate get watcher timestamp

tubi70:
Soundtrack of this PR: [link to song that really fits the mood of this PR]()

### Motivation

Currently the watcher database provides an API for getting block signatures and from there we can get timestamps for a block. However, sometimes the watcher fails or gets stuck. Both fog ingest and fog ledger have code for handling these errors and retries. Because the clients need a timestamp on every piece of data in order to function, failing to get a timestamp blocks the service, and immediately escalates to a P0 issue. We now have the code for getting the timestamps from the watcher in two places — in the fog-ingest worker thread and in the fog ledger worker thread, and specifically a function "get_watcher_timestamp" .https://github.com/mobilecoinfoundation/fog/blob/5a776bd027e1e2cc5f4717f4d4f8f56f30ae0585/fog/ingest/server/src/worker.rs#L151
that is duplicated in both places.

### In this PR
It would be better to have a crate like "fog-timestamps" to provides this function, and reconcile any differences between the two versions, and test this. So that we know it works the same way everywhere, and this will make it easier to debug if fog gets blocked on this in production (which has happened before, and will happen again).

< Ticket status, e.g. "fixes #issue number" > 
https://app.asana.com/0/1200353042931237/1200688378834228/f

### Future Work
* < Out of scope non-goals for this PR >
* < These should be links to tickets. If the tickets do not exist, make them. >

### DIFF
--- a/fog/fog_timestamps/src/watcher.rs
+++ b/fog/fog_timestamps/src/watcher.rs
@@ -1,0 +1,71 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
+
+use mc_common::logger::{log, Logger};
+use mc_transaction_core::BlockIndex;
+use mc_watcher::watcher_db::WatcherDB;
+use mc_watcher_api::TimestampResultCode;
+use std::time::{Duration, Instant};
+
+/// Poll for new data every 10 ms
+const POLLING_FREQUENCY: Duration = Duration::from_millis(10);
+/// If a database invariant is violated, e.g. we get block but not block
+/// contents, it typically will not be fixed and so we won't be able to
+/// proceed. But bringing the server down is costly from ops POV because
+/// we will lose all the user rng's.
+///
+/// So instead, if this happens, we log an error, and retry in 1s.
+/// This avoids logging at > 1hz when there is this error, which would be
+/// very spammy. But the retries are unlikely to eventually lead to
+/// progress. Another strategy might be for the server to enter a
+/// "paused" state and signal for intervention.
+const ERROR_RETRY_FREQUENCY: Duration = Duration::from_millis(1000);
+
+// Get the timestamp from the watcher, or an error code,
+// using retries if the watcher fell behind
+pub fn get_watcher_timestamp(
+    block_index: BlockIndex,
+    watcher: &WatcherDB,
+    watcher_timeout: Duration,
+    logger: &Logger,
+) -> u64 {
+    // Timer that tracks how long we have had WatcherBehind error for,
+    // if this exceeds watcher_timeout, we log a warning.
+    let mut watcher_behind_timer = Instant::now();
+    loop {
+        match watcher.get_block_timestamp(block_index) {
+            Ok((ts, res)) => match res {
+                TimestampResultCode::WatcherBehind => {
+                    if watcher_behind_timer.elapsed() > watcher_timeout {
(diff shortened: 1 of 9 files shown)


eranrund:
Any reason to not have this in the `mobilecoin` repository? This is not fog-specific.
Since this heavily relies on `mc-watcher`, maybe it should live inside that crate?

fog/fog_timestamps/src/watcher.rs
```diff
@@ -0,0 +1,71 @@
...
+) -> u64 {
+    // Timer that tracks how long we have had WatcherBehind error for,
+    // if this exceeds watcher_timeout, we log a warning.
+    let mut watcher_behind_timer = Instant::now();
+    loop {
+        match watcher.get_block_timestamp(block_index) {
+            Ok((ts, res)) => match res {
+                TimestampResultCode::WatcherBehind => {
+                    if watcher_behind_timer.elapsed() > watcher_timeout {
+                        log::warn!(logger, "watcher is still behind on block index = {} after waiting {} seconds, ingest will be blocked", block_index, watcher_timeout.as_secs());
```
eranrund, line 39: `ingest` no longer applies here. Applies to other places below.

fog/fog_timestamps/src/watcher.rs
```diff
@@ -0,0 +1,71 @@
...
+use mc_watcher::watcher_db::WatcherDB;
+use mc_watcher_api::TimestampResultCode;
+use std::time::{Duration, Instant};
+
+/// Poll for new data every 10 ms
+const POLLING_FREQUENCY: Duration = Duration::from_millis(10);
+/// If a database invariant is violated, e.g. we get block but not block
+/// contents, it typically will not be fixed and so we won't be able to
+/// proceed. But bringing the server down is costly from ops POV because
+/// we will lose all the user rng's.
```
eranrund, line 14: user rngs do not apply here.

eranrund (commented):
Some nits, but my main request would be to place this as a method in `WatcherDB`.

garbageslam:
It's not super easy to see from the diff what the differences are here -- can you document what the change is in which server, and what you think the consequences of that might be?

fog/fog_timestamps/src/watcher.rs
tubi70: please let me know if below is the change you are requesting: 

log::warn!(logger, "watcher is still behind on block index = {} after waiting {} seconds, caller will be blocked", block_index, watcher_timeout.as_secs());

fog/fog_timestamps/src/watcher.rs
tubi70: please confirm that below is a better replacement for the comment

/// we will lose all the blocks

eranrund: We will not be losing any blocks. We have no way of knowing what effect broken database invariants has on the caller since we don't know what the caller is doing.
I would change `But bringing the server down is costly from ops POV because
/// we will lose all the user rng's.` to something like `Generally when an invariant is violated we would panic, but this code is used in services that are expensive to restart (such as the ingest enclave).`

fog/fog_timestamps/src/watcher.rs
eranrund: > Any reason to not have this in the `mobilecoin` repository? This is not fog-specific.
> Since this heavily relies on `mc-watcher`, maybe it should live inside that crate?

Ping on this.

PR #124, opened 2021-08-12
Fogdistrotransactionchanges

tubi70:
Soundtrack of this PR: [link to song that really fits the mood of this PR]()

### Motivation

Currently Fog distro is a modified version of slam that sends fog transactions. What causes it to give up is, it builds a bunch of transactions in one thread, Then a sender thread just hits consensus and tries to submit. And if there is an error it retries. However, sometimes the transaction will never succeed, if it was built a while ago with old tombstone block / older fog report. So that would cause distro to fail entirely. So we made distro stop retrying in those cases. But what it should actually do is rebuild the transaction

### In this PR
Add transaction to a queue of transactions to rebuild when distro fails due to old tombstone block

< Ticket status, e.g. "fixes #issue number" > 

### Future Work
* < Out of scope non-goals for this PR >
* < These should be links to tickets. If the tickets do not exist, make them. >

### DIFF
--- a/fog/distribution/src/main.rs
+++ b/fog/distribution/src/main.rs
@@ -278,7 +278,6 @@ fn main() {
     // Spawn worker threads
     for i in 0..config.max_threads {
         let spendable_txouts_receiver2 = spendable_txouts_receiver.clone();
-        let spendable_txouts_sender2 = spendable_txouts_sender.clone();
         let running_threads_sender2 = running_threads_sender.clone();
         let config2 = config.clone();
         let ledger_db2 = ledger_db.clone();
@@ -292,7 +291,6 @@ fn main() {
             .spawn(move || {
                 worker_thread_entry(
                     spendable_txouts_receiver2,
-                    spendable_txouts_sender2,
                     running_threads_sender2,
                     config2,
                     ledger_db2,
@@ -322,7 +320,6 @@ fn main() {
 
 fn worker_thread_entry(
     spendable_txouts_receiver: crossbeam_channel::Receiver<SpendableTxOut>,
-    spendable_txouts_sender: crossbeam_channel::Sender<SpendableTxOut>,
     running_threads_sender: crossbeam_channel::Sender<usize>,
     config: Config,
     ledger_db: LedgerDB,
@@ -358,30 +355,28 @@ fn worker_thread_entry(
 
         // Send to the next fog account
         let to_account = &fog_accounts[txs_created % fog_accounts.len()];
+        // Sometime transactions could not be submitted before tombstone block passed,
+        // so loop until transactions can be submmitted
+        let mut done = false;
+        while !done {
+            // Got our inputs, construct transaction.
+            let tx = build_tx(
+                &pending_spendable_txouts,
+                to_account,
+                &config,
+                &ledger_db,
+                fog_resolver.clone(),
+                &logger,
+            );
 
-        // Got our inputs, construct transaction.
-        let tx = build_tx(
-            &pending_spendable_txouts,
-            to_account,
-            &config,
-            &ledger_db,
-            fog_resolver.clone(),
-            &logger,
-        );
+            txs_created += 1;
 
-        txs_created += 1;
-
-        // Submit tx
-        if submit_tx(
-            txs_created,
-            &conns,
-            &tx,
-            &config,
-            &logger,
-            &spendable_txouts_sender,
-        ) {
-            let mut map = TX_PUB_KEY_TO_ACCOUNT_KEY.lock().unwrap();
-            map.insert(tx.prefix.outputs[0].public_key, to_account.clone());
+            // Submit tx
+            if submit_tx(txs_created, &conns, &tx, &config, &logger) {
+                let mut map = TX_PUB_KEY_TO_ACCOUNT_KEY.lock().unwrap();
+                map.insert(tx.prefix.outputs[0].public_key, to_account.clone());
+                done = true;
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


fog/distribution/src/main.rs
```diff
@@ -369,7 +372,14 @@ fn worker_thread_entry(
...
         // Submit tx
-        if submit_tx(txs_created, &conns, &tx, &config, &logger) {
+        if submit_tx(
+            txs_created,
+            &conns,
+            &tx,
+            &config,
+            &logger,
+            &spendable_txouts_sender,
+        ) {
```
garbageslam, line 382: So I want to suggest another direction:

It looks that `submit_tx` currently returns `true` if it succeeded, and `false` if it fails.
(Ideally this would be documented)

One option would be, if it fails, then this function puts the SpendableTxOut's back in the queue.

And then we don't need the `rebuild_tx` function at all, right?

fog/distribution/src/main.rs
```diff
@@ -355,23 +355,28 @@ fn worker_thread_entry(
 
         // Send to the next fog account
         let to_account = &fog_accounts[txs_created % fog_accounts.len()];
+        // Sometime transactions could not be submitted before tombstone block passed,
+        // so loop until transactions can be submmitted
+        let mut done = false;
+        while !done {
```
garbageslam, line 361: the code can be simpler if this is

```
loop {
...
```

and below, instead of `done = true;` we have `break`

ideally we would have a log message like "rebuilding failed tx" or something so that we know when we are going around this loop

fog/distribution/src/main.rs
```diff
@@ -355,23 +355,28 @@ fn worker_thread_entry(
...
-            &pending_spendable_txouts,
-            to_account,
-            &config,
-            &ledger_db,
-            fog_resolver.clone(),
-            &logger,
-        );
-
-        txs_created += 1;
+            txs_created += 1;
```
garbageslam, line 372: the main point of this counter is that it determines which account is going to recieve the payment

we generally want to make sure all the accounts get funded for testing purposes

so when we rebuild, we don't want to rebuild the transaction for a new recipient

so we should only increment this counter after submit_tx has succeeded

fog/distribution/src/main.rs
```diff
@@ -355,23 +360,58 @@ fn worker_thread_entry(
...
+                    let mr_signer_verifier =
+                        fog_ingest_enclave_measurement::get_mr_signer_verifier(None);
+
+                    let mut verifier = Verifier::default();
+                    verifier.debug(DEBUG_ENCLAVE).mr_signer(mr_signer_verifier);
+                    verifier
+                };
+
+                fog_resolver = FogResolver::new(responses.clone(), &report_verifier)
+                    .expect("Could not get FogResolver");
```
garbageslam, line 412: This looks really good

I think it is slightly better if the report_verifier and Fog resolver::new calls are also in the function you are adding, because then that code won't appear in two places

fog/distribution/src/main.rs
```diff
@@ -318,13 +291,45 @@ fn main() {
     thread::sleep(Duration::from_secs(1));
 }
 
+fn rebuild_fogresolver(
```
garbageslam, line 294: nit: this should be `build_fog_resolver`

and should include the line

```
    let report_verifier = {
        let mr_signer_verifier = fog_ingest_enclave_measurement::get_mr_signer_verifier(None);
        let mut verifier = Verifier::default();
        verifier.debug(DEBUG_ENCLAVE).mr_signer(mr_signer_verifier);
        verifier
    };
```
and the `FogResolver::new` part

fog/distribution/src/main.rs
```diff
@@ -355,23 +363,46 @@ fn worker_thread_entry(
...
-            &pending_spendable_txouts,
-            to_account,
-            &config,
-            &ledger_db,
-            fog_resolver.clone(),
-            &logger,
+        let env = Arc::new(
+            grpcio::EnvBuilder::new()
+                .name_prefix("FogPubkeyResolver-RPC".to_string())
+                .build(),
```
garbageslam, line 369: this `grpcio::EnvBuilder` object contains an entire thread-pool

ideally, there would only be one of these for the whole app, and we pass it to the worker thread, and the `build_fog_resolver` function

i think it would be okay for the `build_fog_resolver` function to also build the `GrpcFogReportConnection`

garbageslam (approved):
looks good thank you!

garbageslam:
lets fix these linter errors and then try deploying

fog/distribution/src/main.rs
```diff
@@ -318,14 +281,57 @@ fn main() {
...
+}
+
 fn worker_thread_entry(
     spendable_txouts_receiver: crossbeam_channel::Receiver<SpendableTxOut>,
     running_threads_sender: crossbeam_channel::Sender<usize>,
     config: Config,
     ledger_db: LedgerDB,
     fog_accounts: Vec<AccountKey>,
-    fog_resolver: FogResolver,
+    mut fog_resolver: FogResolver,
```
eranrund, line 332: Why does this needs to be passed as a parameter if the method can create it on its own? Any reason not to have this function create it before the loop?

garbageslam: i advocated for this to be the current way, the reasoning is, if it happens before `worker_thread_entry`, probably it happens once and all the threads can use it. if it happens always at the start of `worker_thread_entry` then all the threads have to hit the report server at the beginning

it probably doesn't make too much difrerence

sugargoat (approved):

fog/distribution/src/main.rs
eranrund: More load on the report server is not a bad thing to test for :)

eranrund (approved):

fog/distribution/src/main.rs
garbageslam: yeah that's fair

PR #143, opened 2021-12-07
configure sql timeouts

cbeck88:
Brings updates related to SQL timeouts to the release 1.1.2 branch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1714,6 +1714,8 @@ dependencies = [
  "r2d2",
  "rand 0.8.3",
  "rand_core 0.6.2",
+ "serde",
+ "structopt",
  "tempdir",
 ]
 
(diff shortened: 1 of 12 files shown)


jgreat (changes requested):
Timings look good,  Can we get these setting exposed as config values with these setting as default (ideally as env vars, but I'll take cli args, or both).  

We may need to tune and adjust, and it would be nice to not have to cut a release to do that.

cbeck88:
@jgreat let's review this version first, then i'll backport a version of that to 1.1.2, cause it's a larger change to make this configurable everywhere and i'm trying to do it in the cleanest way

https://github.com/mobilecoinfoundation/mobilecoin/pull/1204

wjuan-mob (approved):

cbeck88:
This is now a backport of the PR merged into master for this, I think it should be ready to go

jcape (approved):

jgreat (approved):

PR #145, opened 2021-12-14
backport go-grpc-gateway to release-1.1.2 branch

cbeck88:
In order to do this, we need all proto files that the go-grpc-gateway
    depends on to have a go package annotation. This is why we didn't
    do this before because it relies on changes in the mobilecoin.git
    repo at pretty old commits.
    
    We made those changes on v1.1.0 tag, and pushed a new tag which is
    v1.1.0+, which contains the go-grpc-gateway changes for protos.
    
    This commit
    * makes fog point to that version of mobilecoin
    * renames a proto message in ledger.proto to be called BlockData
      instead of Block to resolve a conflict (as we did in master)
    * updates fog ledger service for the new struct names
    * backports the go-grpc-gateway source and build (but not the tests)
    
    I have confirmed that go-grpc-gateway is building but I have not tested
    it. But since its essentially the same code and protos it should work.
    We can test it with clients in the next step of the release process.


voloshyn (approved):
LGTM

jgreat (approved):

